PR #5, opened 2022-07-27
FFmpeg video compositor pipeline

Janix4000:
Implementation of native FFmpeg in elixir pipeline


test/ffmpeg/ffmpeg_native_test.exs
```diff
@@ -0,0 +1,66 @@
+defmodule VideoCompositor.FFmpeg.Native.Test do
+  use ExUnit.Case, async: true
+
+  alias Membrane.RawVideo
+  alias Membrane.VideoCompositor.FFmpeg.Native
+  alias Membrane.VideoCompositor.Test.Utility
+
+  @tag :tmp_dir
+  test "compose doubled raw video frames on top of each other", %{tmp_dir: tmp_dir} do
```
DominikWolek, line 9: Something's not well configured, because `tmp_dir` is included in git with the output files.

test/test_helper.exs
```diff
@@ -1 +1,5 @@
-ExUnit.start(capture_log: true, exclude: [long: true])
+excluded = [
+  long: true
+]
+
+ExUnit.start(capture_log: true, exclude: excluded)
```
DominikWolek, line 5: This seems unecessary complicated, but we can leave it that way. Also, I commented on previous PR but it wasn't applied: we can just tag tests with `@tag :long` (instead of `@tag long: true`)

test/ffmpeg/ffmpeg_pipeline_test.exs
```diff
@@ -0,0 +1,48 @@
+defmodule Membrane.VideoCompositor.FFmpeg.Pipeline.Test do
```
DominikWolek, line 1: Merge it with `pipeline_integration_test`

lib/frame_composer/ffmpeg_research_compositor_module.ex
```diff
@@ -0,0 +1,41 @@
...
+    %{state: internal_state, iter: iter} = state_of_init_module
+
+    repeated_frames = Stream.repeatedly(fn -> frames.second end) |> Enum.take(div(iter, 80))
+    frames = [frames.first, frames.second] ++ repeated_frames
+
+    internal_state =
+      if rem(iter, 80) == 0 do
+        raw_videos = for _n <- 1..(div(iter, 80) + 2), do: state_of_init_module.raw_video
+
+        {:ok, new_internal_state} = FFmpeg.init(raw_videos)
```
DominikWolek, line 31: NOT: would be nic if you would quickly describe it in the moduledoc - or is it temporary and you are going to change the algorithm in the future?

test/pipeline_integration_test.exs
```diff
@@ -5,39 +5,59 @@ defmodule Membrane.VideoCompositor.PipelineTest do
...
-        first_h264_video_path: "./test/fixtures/short_videos/input_2s_720p.h264",
-        second_h264_video_path: "./test/fixtures/short_videos/input_2s_720p.h264"
+    test "2s 720p 30fps video", %{tmp_dir: tmp_dir} do
+      video_caps = %RawVideo{
+        width: 1280,
+        height: 720,
+        framerate: {30, 1},
+        pixel_format: :I420,
+        aligned: nil
       }
```
DominikWolek, line 20: If you make 1080p and 720p video caps a module attributes, tests could have only one line, something like:
```
test "1s 1080p 30fps video", %{tmp_dir: tmp_dir} do
  test_h264_pipeline(@raw_video_hd_caps, 1, :nx, "short_videos", tmp_dir)
end
```

lib/frame_compositor/ffmpeg_research.ex
```diff
@@ -0,0 +1,47 @@
...
+
+    repeated_frames = Stream.repeatedly(fn -> frames.second end) |> Enum.take(div(iter, 80))
+    frames = [frames.first, frames.second] ++ repeated_frames
+
+    internal_state =
+      if rem(iter, 80) == 0 do
+        raw_videos = for _n <- 1..(div(iter, 80) + 2), do: state_of_init_module.raw_video
+
+        {:ok, new_internal_state} = FFmpeg.init(raw_videos)
+        {:ok, new_internal_state} = FFmpeg.duplicate_metadata(new_internal_state, internal_state)
```
DominikWolek, line 38: Maybe we could have a init function with two arguments
```
spec init(videos :: [raw_video], last_pts :: int)
```
that would make `duplicate_metadata` redundant

DominikWolek (changes requested):
Also, remove the `test/fixtures/tmp_dir` dir

lib/frame_compositor/ffmpeg_research.ex
Janix4000: More likely `spec init(videos :: [raw_video], last_state :: state)`.
There might be more fields to copy in the future (even separate pts for every video). I would leave it as it is, right now.

test/pipeline_integration_test.exs
Janix4000: I'll change this in PR#9

DominikWolek (approved):

PR #12, opened 2022-08-18
Rust-based OpenGL compositor

jerzywilczek:
This PR adds the Rust-based OpenGL compositor implementation. 
The implementation has many flaws. If we omit the lacking functionality (it has feature-parity with the C++ version, but that's hardly much), the most serious flaw is the lack of proper error handling. I however feel this should be merged as soon as I finish writing the docs.


mkaput (changes requested):

native/membrane_videocompositor_opengl_rust/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "membrane_videocompositor_opengl_rust"
+version = "0.1.0"
+authors = []
+edition = "2018"
```
mkaput, line 5:
```suggestion
edition = "2021"
```

native/membrane_videocompositor_opengl_rust/src/framebuffers.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+#[repr(usize)]
+#[derive(Debug, Clone, Copy)]
+/// Represents a plane in a YUV planar image format.
+pub enum Plane {
+  Y = 0,
+  U,
+  V,
+}
```
mkaput, line 171:
```suggestion
/// Represents a plane in a YUV planar image format.
#[repr(usize)]
#[derive(Debug, Clone, Copy)]
pub enum Plane {
  Y = 0,
  U,
  V,
}
```

this order is more natural

native/membrane_videocompositor_opengl_rust/src/framebuffers.rs
```diff
@@ -0,0 +1,171 @@
...
+
+  /// Get the width of the Y plane.
+  pub fn width(&self) -> usize {
+    self.width 
+  }
+
+  /// Get the height of the Y plane.
+  pub fn height(&self) -> usize {
+    self.height 
+  }
```
mkaput, line 161: don't do getters/setters unless it's really needed :) I think you should be easily able to make these fields simply `pub`

native/membrane_videocompositor_opengl_rust/src/framebuffers.rs
```diff
@@ -0,0 +1,171 @@
...
+    assert!(buffer.len() >= pixels_amount * 3 / 2);
+
+    unsafe {
+      self.framebuffers[0].read_to_ptr(buffer.as_mut_ptr());
+
+      self.framebuffers[1].read_to_ptr(buffer.as_mut_ptr().add(pixels_amount));
+
+      self.framebuffers[2].read_to_ptr(buffer.as_mut_ptr().add(pixels_amount * 5 / 4));
+    }
+  }
```
mkaput, line 151: consider implementing this trait instead https://doc.rust-lang.org/std/io/trait.Read.html

native/membrane_videocompositor_opengl_rust/src/framebuffers.rs
```diff
@@ -0,0 +1,171 @@
...
+  /// Create a new instance.
+  ///  * `width` and `height` should be given in pixels.
+  ///  * `internal_format` is the format used by the renderbuffer internally (e.g `gl::RGB8` or `gl::R8`)
+  ///  * `format` and `type_` represent the type and format we expect to get out when reading from the framebuffer (e.g. `gl::RGB` for `format` and `gl::UNSIGNED_BYTE` for `type_`)
+  fn new(
+    width: usize,
+    height: usize,
+    internal_format: gl::GLenum,
+    format: gl::GLenum,
+    type_: gl::GLenum,
```
mkaput, line 25:
```suggestion
    object_type: gl::GLenum,
```

or something similar; should end up more readable

mkaput (changes requested):

mkaput (commented):
Have you considered using https://github.com/gfx-rs/wgpu instead of OpenGL directly?

native/membrane_videocompositor_opengl_rust/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+    egl_bad_surface,
+    egl_bad_match,
+    egl_bad_parameter,
+    egl_bad_native_pixmap,
+    egl_bad_native_window,
+    egl_context_lost,
+  }
+}
+
+#[derive(Debug, rustler::NifStruct)]
```
mkaput, line 40: FYI that works

```suggestion
use rustler::NifStruct;

#[derive(Debug, NifStruct)]
```

native/membrane_videocompositor_opengl_rust/glad-gles2/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
```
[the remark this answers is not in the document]
jerzywilczek: it isn't. this is generated by [GLAD2](https://gen.glad.sh/), and sadly it's the only place you can get this package, and there are no good OpenGL ES crates on crates.io ;(

native/membrane_videocompositor_opengl_rust/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+
+
+  egl.make_current(display, None, None, None).nif_err()?;
+  Ok((atoms::ok(), ResourceArc::new(State::new(display, context, scene))))
+}
+
+#[rustler::nif]
+/// Join two frames passed as [binaries](rustler::Binary). This function is intended to only be called by Elixir.
+fn join_frames<'a>(env: rustler::Env<'a>, state: rustler::ResourceArc<State>, upper: rustler::Binary, lower: rustler::Binary) -> Result<(rustler::Atom, rustler::Term<'a>), rustler::Error> {
+  let ctx: BoundContext = state.bind_context().nif_err()?;
```
mkaput, line 192: type annotation can be elided

```suggestion
  let ctx = state.bind_context().nif_err()?;
```

native/membrane_videocompositor_opengl_rust/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+
+trait ResultExt<T> {
+  /// Convert `T` into a [Result] with a [rustler::Error].
+  fn nif_err(self) -> Result<T, rustler::Error>;
+}
+
+impl<T> ResultExt<T> for Result<T, egl::Error> {
+  fn nif_err(self) -> Result<T, rustler::Error> {
+    use rustler::Error;
+    match self {
```
mkaput, line 218: https://doc.rust-lang.org/std/result/enum.Result.html#method.map_err

native/membrane_videocompositor_opengl_rust/rustfmt.toml
```diff
@@ -0,0 +1 @@
+tab_spaces = 2
```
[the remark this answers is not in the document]
jerzywilczek: I know, but Membrane uses 2-space indentation everywhere, even for C/C++ code, and I figured I should set it here as well.
@DominikWolek can I use 4-space indentation for Rust code?

native/membrane_videocompositor_opengl_rust/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+  let mut binary = rustler::OwnedBinary::new(ctx.scene.out_width() * ctx.scene.out_height() * 3 / 2).unwrap();
+  ctx.scene.draw_into(binary.as_mut_slice());
+  Ok((atoms::ok(), binary.release(env).to_term(env)))
+}
+
+
+trait ResultExt<T> {
+  /// Convert `T` into a [Result] with a [rustler::Error].
+  fn nif_err(self) -> Result<T, rustler::Error>;
+}
```
mkaput, line 213: That's smart, but actually I think I would generalise this over any Error-like types, instead of being focused just on `Result<T, E>` wrapper.

If we hadn't have orphan rule in Rust, you could implement `impl From<egl::Error> for rustler::Error`. For now, you can define a `From`-like trait which is `rustler::Error` specific and put impls for all interesting structs.

mkaput (changes requested):

native/membrane_videocompositor_opengl_rust/build.rs
```diff
@@ -0,0 +1,5 @@
+fn main() {
+  println!("cargo:rustc-link-lib=EGL");
+  println!("cargo:rustc-link-lib=GLESv2");
+  println!("cargo:rustc-link-search=../..");
```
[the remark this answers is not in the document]
jerzywilczek: MacOS doesn't have EGL, which we need to use for this project. We use [Google ANGLE](https://github.com/google/angle) on MacOS, and we ended up just putting the two libraries in the project root if we're compiling on MacOS. That was back when we were using C++ for the prototype and we had to do it because of Unifex weirdness on MacOS. When we started using Rust, the libraries were already established to be there.

native/membrane_videocompositor_opengl_rust/glad-gles2/src/lib.rs
```diff
@@ -0,0 +1,2 @@
+#[allow(clippy::all)]
+pub mod gl;
```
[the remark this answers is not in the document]
jerzywilczek: This isn't my code. Should I correct it anyway, even though this is a different crate?

README.md
```diff
@@ -17,6 +17,9 @@ def deps do
   ]
 end
 ```
+
+Since parts of this package are implemented in Rust, you need to have a Rust installation to compile this package. You can get one [here](https://rustup.rs/)
```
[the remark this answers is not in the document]
jerzywilczek: I'm skeptical whether this would work, since every computer can have a different OpenGL implementation and we need to link against it. Doesn't seem like something that could realistically be done, since every graphics driver update may potentially require a relink.

native/membrane_videocompositor_opengl_rust/src/lib.rs
jerzywilczek: You mean I should do something like this:

```rs
trait ToRustlerError {
    fn convert(self) -> rustler::Error;
}

impl ToRustlerError for egl::Error {
    ...
}
```
and then `.map_err(|e| e.convert())` everywhere?

native/membrane_videocompositor_opengl_rust/src/framebuffers.rs
jerzywilczek: I don't really like that, since `std::io::Read` automatically provides many methods that wouldn't make sense here, like `read_exact` (data in this buffer doesn't make sense split into parts) or `read_to string` (this is an image, certainly wouldn't be valid UTF-8). My understanding of it is that it's not really meant for binary data. Am I wrong here?

native/membrane_videocompositor_opengl_rust/src/framebuffers.rs
jerzywilczek: I don't like other modules being able to modify these fields, since the safety of the `read` method depends on them being correct. Should I really do it?

native/membrane_videocompositor_opengl_rust/src/lib.rs
jerzywilczek: I know, I just like leaving a 'namespace' in places like this, rather than having a use for the entire module.

jerzywilczek:
@mkaput We did consider wgpu, but it was rejected as a backend candidate earlier in the project. I was actually the one to suggest researching it.

native/membrane_videocompositor_opengl_rust/src/lib.rs
jerzywilczek: Actually, I think I came up with a potentially better solution while working on error handling. I can have my own `CompositorError` type, which implements `From<egl::Error>` (along with a couple more `From<T>`s for all kinds of errors that can arise in the project) and implement `From<CompositorError>` for `rustler::Error`. Then error handling becomes very nice, since we can just use the `?` operator virtually everywhere we want. What do you think of that?

README.md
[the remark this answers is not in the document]
mkaput: isn't opengl implementation provided by the drivers linked dynamically when you initialize gl?

native/membrane_videocompositor_opengl_rust/glad-gles2/src/lib.rs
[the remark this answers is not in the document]
mkaput: ok, I wasn't aware of this.

native/membrane_videocompositor_opengl_rust/glad-gles2/Cargo.toml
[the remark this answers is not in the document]
mkaput: yeah, there may be no opengl es specific because the community is much more focused on WebGPU which works better with Rust

mkaput, line 1: it would be nice to put w README here documenting how to regenerate the bindings

mkaput (commented):
native/membrane_videocompositor_opengl_rust/rustfmt.toml can be deleted

native/membrane_videocompositor_opengl_rust/src/lib.rs
mkaput: do as you wish :) this was just a hint

native/membrane_videocompositor_opengl_rust/src/lib.rs
mkaput: Yup, that sounds good!

native/membrane_videocompositor_opengl_rust/src/shaders.rs
```diff
@@ -0,0 +1,109 @@
...
+    CantLinkProgram,
+}
+
+mod atoms {
+    rustler::atoms! {
+      cant_compile_fragment_shader,
+      cant_compile_vertex_shader,
+      cant_link_program
+    }
+}
```
mkaput, line 93: I wonder if it wouldn't be easier if you just used regular strings/binaries for error messages. You could use the `thiserror` crate to nicely build error enum with nice Display implementation and then much more shortly build rustler terms out of this.

mkaput: also I would avoid using English informal shortcuts in API boundaries

native/membrane_videocompositor_opengl_rust/src/scene.rs
```diff
@@ -0,0 +1,252 @@
...
+    framebuffers::{self, YUVRenderTarget},
+    shaders::ShaderProgram,
+    textures::YUVPlanarTexture,
+};
+use glad_gles2::gl;
+
+/// Describes where a video should be located in the scene space.
+/// All coordinates have to be in the range [-1, 1].
+pub struct VideoPlacementTemplate {
+    pub top_right: (f32, f32),
```
mkaput, line 13:
```suggestion
    pub top_right: Vec2f,
```

I would consider introducing a struct for this tuple. Maybe even using `nalgebra` provided types, but these will probably increase compilation time a lot.

native/membrane_videocompositor_opengl_rust/src/framebuffers.rs
mkaput: ok, I didn't notice that dependency. I think you should be fine with getters here. There is a crate called `readonly`, but I think it's more of a fun hack than production quality approach.

native/membrane_videocompositor_opengl_rust/src/framebuffers.rs
```diff
@@ -0,0 +1,171 @@
...
+        }
+    }
+
+    fn bind_for_reading(&self) {
+        unsafe {
+            gl::BindFramebuffer(gl::READ_FRAMEBUFFER, self.id);
+            gl::ReadBuffer(gl::COLOR_ATTACHMENT0);
+        }
+    }
+
```
mkaput, line 76: this API seems very unsafe. I don't know opengl, but it looks like you have to bind before drawing into. it would be more idiomatic then to return some kind of a proxy structure here that would expose the draw/read methods and keep track of the binding status.

mkaput (changes requested):
Is this code thread-safe? because the structures are Send + Sync which I think isn't true here

native/membrane_videocompositor_opengl_rust/src/framebuffers.rs
mkaput: hmm, this makes sense. I wonder though if you don't have generic behaviours called `reading` and `drawing` (aren't they the same thing though?).

Maybe it would make sense to define something like `trait Readable`

README.md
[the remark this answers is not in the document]
jerzywilczek: I honestly don't know. I've been having linking issues for a long time in the experimental stage of this project. It should be possible to link dynamically, but I think this could be a lot of work and might be out of the scope of this PR. Do you think it's OK to merge this as-is and add figuring out precompilation to the backlog?

native/membrane_videocompositor_opengl_rust/src/lib.rs
jerzywilczek: I'll resolve this conversation then, since I already have a PR implementing this idea.

README.md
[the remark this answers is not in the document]
mkaput: yes, that's definitely ok to do this in separate PR :)

native/membrane_videocompositor_opengl_rust/glad-gles2/Cargo.toml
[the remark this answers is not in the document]
jerzywilczek: We just talked about this in our daily. My next task is re-researching and reconsidering wgpu as a backend candidate!

native/membrane_videocompositor_opengl_rust/src/framebuffers.rs
jerzywilczek: I think there is no unsafety here, since OpenGL should just throw an error when trying to draw without binding, but the idea of a proxy makes a lot of sense here.

native/membrane_videocompositor_opengl_rust/src/scene.rs
jerzywilczek: We will probably use some OpenGL-specific maths library, which would provide functionality similar to `nalgebra` in the future. For now I'll just make a Point(f32, f32)

native/membrane_videocompositor_opengl_rust/src/framebuffers.rs
jerzywilczek: I think this may make sense later in the project, when we have more render target types (e.g for rendering RGB). Can I leave it as-is since I don't know which behaviors will be shared with other render targets and do this when I know more about this?

native/membrane_videocompositor_opengl_rust/src/shaders.rs
jerzywilczek: I think this can be addressed in #13 instead of here maybe? I added a comment in there linking to this conversation.

jerzywilczek:
@mkaput You are right that the structures shouldn't be `Send` or `Sync`. How should I unimplement these? Use `#![feature(negative_impls)]` and `impl !Send` or put a dummy raw pointer inside them, or maybe something else entirely? I've never done this sort of thing and don't know what would be more idiomatic.

native/membrane_videocompositor_opengl_rust/glad-gles2/Cargo.toml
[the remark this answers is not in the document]
mkaput: great!

native/membrane_videocompositor_opengl_rust/src/framebuffers.rs
mkaput: sure, you can leave this for now

native/membrane_videocompositor_opengl_rust/src/shaders.rs
mkaput: 👍🏻

native/membrane_videocompositor_opengl_rust/src/scene.rs
mkaput: 👍🏻

native/membrane_videocompositor_opengl_rust/src/framebuffers.rs
mkaput: how is the error captured though? 😄 won't this just panic the whole machine? even if this is a programmer's error, it would be nice if abstractions prevented writing bad code

mkaput:
> @mkaput You are right that the structures shouldn't be `Send` or `Sync`. How should I unimplement these? Use `#![feature(negative_impls)]` and `impl !Send` or put a dummy raw pointer inside them, or maybe something else entirely? I've never done this sort of thing and don't know what would be more idiomatic.

https://stackoverflow.com/questions/62713667/how-to-implement-send-or-sync-for-a-type

mkaput (approved):
👏🏻 there are some rough edges, but you are working on these, so this one gets my seal of approval

native/membrane_videocompositor_opengl_rust/src/framebuffers.rs
jerzywilczek: it won't. you have to call gl::GetError() to check whether errors have occurred. If you don't react to the error the state machine will be in a state you don't expect and may produce weird results, but it won't crash/panic

PR #13, opened 2022-08-23
Improve our OpenGL error-handling infrastructure.

jerzywilczek:
This patch adds 2 things to the codebase:

 * a `gl` macro, which can be used when calling OpenGL functions. It
   calls an OpenGL function, does error-checking and returns an appropriate `Result` with a very nice (in my opinion) error.
   From now on, this macro should be the only way to call OpenGL.
 * some error-handling types: `GLError` for OpenGL errors and
   `CompositorError` for converting from various errors to the final
   `rustler::Error` in a convenient way.

Based on #12

### DIFF
--- a/native/membrane_videocompositor_opengl_rust/src/errors.rs
+++ b/native/membrane_videocompositor_opengl_rust/src/errors.rs
@@ -13,8 +13,7 @@ pub fn result_or_gl_error<T>(
     res: T,
     file: &str,
     line: u32,
-    function: &str,
-    args: &str,
+    call: &str,
 ) -> Result<T, CompositorError> {
     use glad_gles2::gl;
 
@@ -27,7 +26,7 @@ pub fn result_or_gl_error<T>(
     let location = ErrorLocation {
         file: file.to_string(),
         line,
-        call: format!("gl{}({})", function, args),
+        call: call.to_string(),
     };
 
     match err {
--- a/native/membrane_videocompositor_opengl_rust/src/errors.rs
+++ b/native/membrane_videocompositor_opengl_rust/src/errors.rs
@@ -1,4 +1,3 @@
-use crate::gl;
 use thiserror::Error;
 
 #[derive(Debug, rustler::NifStruct)]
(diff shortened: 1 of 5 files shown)


DominikWolek (approved):

mkaput (commented):
Frankly, I would not recommend using macros here, because not only they will increase compilation time sizeably, they are also breaking any IDE support you have.

I know that knowing which line of code raised an error is nice, but is it worth breaking your editor?

native/membrane_videocompositor_opengl_rust/src/errors.rs
```diff
@@ -0,0 +1,64 @@
...
+    match err {
+        gl!(NO_ERROR) => Ok(res),
+        gl!(INVALID_ENUM) => Err(CompositorError::GLError("gl_invalid_enum", location)),
+        gl!(INVALID_VALUE) => Err(CompositorError::GLError("gl_invalid_value", location)),
+        gl!(INVALID_OPERATION) => Err(CompositorError::GLError("gl_invalid_opegation", location)),
+        gl!(INVALID_FRAMEBUFFER_OPERATION) => Err(CompositorError::GLError(
+            "gl_invalid_framebuffer_operations",
+            location,
+        )),
+        gl!(OUT_OF_MEMORY) => Err(CompositorError::GLError("gl_out_of_memory", location)),
```
mkaput, line 27: what does these `gl!` expand into?

jerzywilczek: That's the second case in the macro
```rs
($name:ident) => {
        glad_gles2::gl::$name
    }
```

jerzywilczek:
@mkaput 
I coded this macro as an experiment, because I was expecting it to break IDE support completely. To my surprise, rust-analyzer in VS Code doesn't break. It still suggests everything correctly, even manages to figure out this is a function call and tells you arguments and their types. Will the compilation times be much longer even with a rather simple macro like this one?

native/membrane_videocompositor_opengl_rust/src/errors.rs
jerzywilczek: so it expands to `glad_gles2::gl::INVALID_ENUM`

mkaput: ok, gotcha; though this is quite not obvious and I would rather avoid such shortcuts in code

mkaput:
> @mkaput I coded this macro as an experiment, because I was expecting it to break IDE support completely. To my surprise, rust-analyzer in VS Code doesn't break. It still suggests everything correctly, even manages to figure out this is a function call and tells you arguments and their types.

Wow, I'm out of the loop totally here.

> Will the compilation times be much longer even with a rather simple macro like this one?

It shouldn't be so much decrease, but I am unable to guess.

---

I wonder if this macro wouldn't be more ergonomic/readable if it didn't glue the `gl::` module path to the expressions and simply did this:

```rust
gl!(gl::ClearColor(...))
```

→

```rust
{
    let result = $expr;
    crate::errors::result_or_gl_error(result, file!(), line!())
}
```

Notice that I have moved all logic into `result_or_gl_error` function in order to minimize the size of the code generated by the macro (this is an important factor when metaprogramming).

jerzywilczek:
I think there is absolutely some logic in the macro that should be moved into the `result_or_gl_error` function. Removing the second case from the macro should simplify it further, but I like the ergonomics of glueing `gl::` in the front in the macro, especially since the IDE support doesn't suffer from it.

mkaput:
> I think there is absolutely some logic in the macro that should be moved into the `result_or_gl_error` function. Removing the second case from the macro should simplify it further, but I like the ergonomics of glueing `gl::` in the front in the macro, especially since the IDE support doesn't suffer from it. Do you think this is OK now, or should I make adding the `gl::` module path necessary?

do whatever suits you :)

native/membrane_videocompositor_opengl_rust/src/lib.rs
```diff
@@ -16,19 +16,13 @@ macro_rules! gl {
         {
             // FIXME: we should probably add something like `ensure_current_thread_holds_context` here
             let result = glad_gles2::gl::$call($($args),*);
-            #[allow(unused_unsafe)] // we have this, because maybe there are some gl calls that are safe and in that case we need our own unsafe block here
-            let err = unsafe { glad_gles2::gl::GetError() };
-            crate::errors::result_or_gl_error(result, err, crate::errors::ErrorLocation {
+            crate::errors::result_or_gl_error(result, crate::errors::ErrorLocation {
                 file: file!().to_string(),
```
mkaput, line 20: I would try to push `.to_string()` and `format!` calls into `result_or_gl_error`, especially `format!` contributes to code size a lot.

mkaput (approved):

jerzywilczek:
Actually I don't think we should merge this right now.

@mkaput I tried using it just now and it turns out removing the second case from the macro (the one that allowed referencing constants like this: `gl!(CONSTANT_NAME)`) does break my IDE, which was apparently thinking the macro would resolve into the second case most of the time. Seeing that this breaks IDE support completely I will implement your idea with passing a single `expr` into the macro. You were right all along!

mkaput (approved):
Looks good, though as discussed in the office, maybe we'll be able to reuse the `track_caller` API here.

PR #14, opened 2022-08-23
Improve error handling for shaders.

jerzywilczek:
* Make all OpenGL calls in the `shaders` module use the new `gl`
   macro and bubble-up the errors.
 * Include the compilation/linking error log downloaded from the GPU in the `ShaderError`.

Based on #13


jerzywilczek:
I think [this](https://github.com/membraneframework-labs/membrane_video_compositor_plugin/pull/12#discussion_r952348401) should be handled in this PR, since it overhauls the shader error handling code

PR #20, opened 2022-08-26
Tests on new docker_membrane with rust

jerzywilczek:
This makes our CI able to compile rust.


piotrekzyla (approved):
😉

piotrekzyla:
I was just told that about the orbs and it is a "new way of doing" it (at first I thought it could be legacy... my bad), so I guess this PR could be scratched. The changes I made to the `membraneframeworklabs/docker_membrane:latest` image should help here anyway.

PR #24, opened 2022-08-30
Multiple input Video Compositor

Janix4000:
Elixir implementation of MI VC. Provides behaviour for frame compositors and tests compositor using mock object.


lib/membrane_video_compositor_multiple_inputs.ex
```diff
@@ -0,0 +1,269 @@
...
+          pads_to_ids: {_new_id, pads_to_ids}
+        } = state
+      ) do
+    id = Map.get(pads_to_ids, pad)
+
+    ids_to_tracks = push_frame(ids_to_tracks, id, buffer)
+
+    state = %{state | ids_to_tracks: ids_to_tracks}
+
+    if all_has_frame?(ids_to_tracks) do
```
jerzywilczek, line 103: This should be `all_have_frame`

lib/membrane_video_compositor_multiple_inputs.ex
```diff
@@ -0,0 +1,294 @@
...
+
+  @impl true
+  def handle_process(
+        pad,
+        buffer,
+        _context,
+        %{
+          ids_to_tracks: ids_to_tracks,
+          pads_to_ids: {_new_id, pads_to_ids}
+        } = state
```
mat-hek, line 129: Don't unpack arguments in the function head unless required to match. Unpack in the function body instead.

lib/membrane_video_compositor_multiple_inputs.ex
```diff
@@ -0,0 +1,294 @@
...
+  defp all_have_frame?(ids_to_tracks) do
+    ids_to_tracks
+    |> Map.values()
+    |> Enum.all?(&Track.has_frame?/1)
+  end
+
+  defp get_ids_to_frames(ids_to_tracks) do
+    ids_to_tracks
+    |> Enum.map(fn {id, %Track{} = track} -> {id, Track.first_frame(track)} end)
+    |> Map.new()
```
mat-hek, line 193:
```suggestion
    |> Map.new(fn {id, %Track{} = track} -> {id, Track.first_frame(track)} end)
```

lib/membrane_video_compositor_multiple_inputs.ex
```diff
@@ -0,0 +1,294 @@
...
+
+      {:merged, :still_ongoing} ->
+        {{:ok, buffer: {:output, buffers}}, state}
+
+      {:merged, :all_finished} ->
+        {{:ok, buffer: {:output, buffers}, end_of_stream: :output}, state}
+    end
+  end
+
+  @dialyzer {:nowarn_function, determine_compositor_module: 1}
```
mat-hek, line 281: maybe `Implementation.get_implementation_module` could be excluded instead?

lib/membrane_video_compositor_multiple_inputs.ex
```diff
@@ -0,0 +1,294 @@
...
+
+  defp push_frame(ids_to_tracks, id, frame) do
+    Map.update!(ids_to_tracks, id, fn track ->
+      Track.push_frame(track, frame)
+    end)
+  end
+
+  defp all_have_frame?(ids_to_tracks) when map_size(ids_to_tracks) == 0 do
+    false
+  end
```
mat-hek, line 182: redundant

lib/membrane_video_compositor_multiple_inputs.ex
```diff
@@ -0,0 +1,294 @@
+defmodule Membrane.VideoCompositor.MultipleInputs do
```
mat-hek, line 1: this should be in the `video_compositor` directory

lib/multiple_inputs_frame_compositor.ex
```diff
@@ -0,0 +1,33 @@
+defmodule Membrane.VideoCompositor.FrameCompositor.MultipleInputs do
```
mat-hek, line 1: this should be in the `video_compositor` directory

mat-hek (changes requested):

lib/membrane_video_compositor_multiple_inputs.ex
```diff
@@ -0,0 +1,294 @@
...
+  The element responsible for placing the first received frame
+  above the other and sending forward buffer with
+  merged frame binary in the payload.
+  """
+
+  use Membrane.Filter
+  alias Membrane.RawVideo
+  alias Membrane.VideoCompositor.MultipleInputs.Implementation
+
+  def_options(
```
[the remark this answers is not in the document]
Janix4000: It is already added

lib/membrane_video_compositor_multiple_inputs.ex
Janix4000: Not exactly. From Elixir's documentation:
```
As the last example shows, Enum.all?/2 returns true if enumerable is empty,
regardless of fun. In an empty enumerable there is no element for which fun
returns a falsy value, so the result must be true. This is a well-defined
logical argument for empty collections.
```
tl;dr, `Enum.all?/2` will return `true` value for empty collection, but in our case, we want `false` result, if there are no tracks in the collection

lib/membrane_video_compositor_multiple_inputs.ex
Janix4000: It is a temporary solution. I leave it as it is.

lib/membrane_video_compositor_multiple_inputs.ex
mat-hek: Hmm, so the name is kind of counterintuitive. Maybe do that check where the function is called

lib/membrane_video_compositor_multiple_inputs.ex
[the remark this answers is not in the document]
mat-hek: So the braces have to be removed manually. The formatter won't touch them after the config is changed

lib/membrane_video_compositor_multiple_inputs.ex
Janix4000: Renamed

lib/membrane/video_compositor/multiple_inputs/video_compositor/implementation.ex
```diff
@@ -0,0 +1,51 @@
+defmodule Membrane.VideoCompositor.MultipleInputs.VideoCompositor.Implementation do
+  @moduledoc """
+  A module describing multiple input video compositor implementation type and implementing
+  functions related with implementation format.
+  """
+
+  @typedoc "Define video compositor implementation types"
+  @type implementation_t :: :ffmpeg | :opengl_cpp | :opengl_rust | :nx | module()
```
mat-hek, line 8: it seems `module()` is not supported

lib/membrane/video_compositor/multiple_inputs/video_compositor/implementation.ex
```diff
@@ -0,0 +1,51 @@
...
+      _other ->
+        {:error, "Format not supported"}
+    end
+  end
+
+  @spec get_test_implementations() :: list()
+  def get_test_implementations() do
+    test_implementations = [:ffmpeg, :nx]
+    test_implementations
+  end
```
mat-hek, line 34: what is it for?

lib/membrane/video_compositor/multiple_inputs/video_compositor/implementation.ex
```diff
@@ -0,0 +1,51 @@
+defmodule Membrane.VideoCompositor.MultipleInputs.VideoCompositor.Implementation do
```
mat-hek, line 1: - `VideoCompositor.MultipleInputs.VideoCompositor` 🤔 I'd generally try to shorten module names
- Isn't it supposed to replace the current compositor?
- I think this should be in a separate PR

lib/membrane/video_compositor/multiple_inputs/video_compositor.ex
```diff
@@ -177,20 +163,19 @@ defmodule Membrane.VideoCompositor.MultipleInputs do
     end)
   end
 
-  defp all_have_frame?(ids_to_tracks) when map_size(ids_to_tracks) == 0 do
+  defp all_tracks_are_ready_to_merge?(ids_to_tracks) when map_size(ids_to_tracks) == 0 do
```
mat-hek, line 166:
```suggestion
  defp all_tracks_ready_to_merge?(ids_to_tracks) when map_size(ids_to_tracks) == 0 do
```

lib/membrane/video_compositor/multiple_inputs/video_compositor.ex
```diff
@@ -87,47 +51,69 @@ defmodule Membrane.VideoCompositor.MultipleInputs do
       caps: options.caps,
       compositor_module: compositor_module,
       internal_state: internal_state,
-      pads_to_ids: {0, %{}}
+      pads_to_ids: %{},
+      new_pad_id: 0,
+      ids_to_positions: %{}
```
mat-hek, line 56: couldn't position be a part of the track?

lib/membrane/video_compositor/multiple_inputs/video_compositor/track.ex
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  @spec push_frame(__MODULE__.t(), buffer_t) :: __MODULE__.t()
+  def push_frame(%__MODULE__{buffers: buffers} = track, frame) do
+    %__MODULE__{track | buffers: Qex.push(buffers, frame)}
+  end
+
+  @spec pop_frame(__MODULE__.t()) :: __MODULE__.t()
+  def pop_frame(%__MODULE__{buffers: buffers} = track) do
+    %__MODULE__{track | buffers: Qex.pop!(buffers) |> elem(1)}
```
mat-hek, line 27: Avoid using `elem`, it's hard to read

lib/membrane/video_compositor/implementation/ffmpeg.ex
```diff
@@ -1,11 +1,11 @@
-defmodule Membrane.VideoCompositor.FFmpeg do
+defmodule Membrane.VideoCompositor.Implementation.FFmpeg do
```
mat-hek, line 1:
```suggestion
defmodule Membrane.VideoCompositor.Implementations.FFmpeg do
```

We usually use the plural for such things, as singular is useful for things like `Implementation.Utils`. But do as you prefer

lib/membrane/video_compositor/multiple_inputs/video_compositor/track.ex
```diff
@@ -0,0 +1,39 @@
+defmodule Membrane.VideoCompositor.MultipleInputs.VideoCompositor.Track do
+  @moduledoc false
+
+  @type buffer_t :: Membrane.Buffer.t()
```
mat-hek, line 4: better alias Membrane.Buffer and use Buffer.t

mat-hek (changes requested):

lib/membrane/video_compositor/multiple_inputs/video_compositor/implementation.ex
Janix4000: - See the following subsection. I will change in the future. I will leave it as it is, right now.
 - We decided to separate these two approaches. In future PRs, we may remove one of them or merge them.
 -  I use it in the following PR, implementing Rust OpenGL, I think it should stay.

lib/membrane/video_compositor/multiple_inputs/video_compositor/implementation.ex
Janix4000: It is visible by the IDE and recognised as a built-in alias for `atom()`

lib/membrane/video_compositor/multiple_inputs/video_compositor/implementation.ex
Janix4000: I removed it

lib/membrane/video_compositor/multiple_inputs/video_compositor/implementation.ex
Janix4000: I misunderstood, removed `module()` from the `@type`

mat-hek (approved):

PR #26, opened 2022-09-05
Add a basic skeleton for the `wgpu`-based video compositor.

jerzywilczek:


native/membrane_videocompositor_wgpu/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "membrane_videocompositor_wgpu"
+version = "0.1.0"
+authors = []
```
DominikWolek, line 4: Let's add Apache licence and SWM as the authors - this repo is public, after all. 
I think we should also apply this to the OpenGL cargo package

DominikWolek (changes requested):

DominikWolek (approved):

PR #29, opened 2022-09-06
Rust opengl multiple input pads Compositor

Janix4000:


native/membrane_videocompositor_opengl_rust/src/lib.rs
```diff
@@ -235,31 +213,156 @@ fn init(
...
+        input_video.width,
+        input_video.height,
+    )?;
+
+    Ok(atoms::ok())
+}
+
+fn lerp(x: f32, from_min: f32, from_max: f32, to_min: f32, to_max: f32) -> f32 {
+    (x - from_min) / (from_max - from_min) * (to_max - to_min) + to_min
+}
```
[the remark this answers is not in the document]
jerzywilczek: also, I find the names of the parameters very confusing. Wouldn't it sound more natural if these were called `x_min`, `x_max`, `y_min`, `y_max`?

native/membrane_videocompositor_opengl_rust/src/scene.rs
```diff
@@ -211,31 +213,40 @@ impl Scene {
...
     /// Upload a texture to a video specified by the `video_idx`
     ///
     /// # Panics
     ///
     /// This function will panic if the `video_idx` is not an index of an existing video and if `data` doesn't have proper length for a YUV420p-encoded frame for the specified video.
-    pub fn upload_texture(&self, video_idx: usize, data: &[u8]) -> Result<(), CompositorError> {
+    pub fn upload_texture(&self, video_idx: &usize, data: &[u8]) -> Result<(), CompositorError> {
         self.videos[video_idx].upload_texture(data)?;
         Ok(())
     }
```
jerzywilczek, line 252: I really dislike the panics here, since a panic will (I believe) crash the BEAM entirely. I find it more acceptable in `draw_into`, since it only operates on objects created and validated internally in this NIF. Here we crash when a value passed down from elixir is incorrect, which I find more bad. Consider making a new `CompositorError` for passing an incorrect video index and returning it from here.

Also, we should later replace panics with errors everywhere later.

jerzywilczek: also, passing a `usize` by reference instead of by value seems unnecessary.

native/membrane_videocompositor_opengl_rust/src/lib.rs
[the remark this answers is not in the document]
Janix4000: Also documented the function

native/membrane_videocompositor_opengl_rust/src/lib.rs
```diff
@@ -199,34 +200,11 @@ fn init(
 
     unsafe { gl!(glad_gles2::gl::ClearColor(0.0, 0.0, 0.0, 1.0))? }
 
-    let vertex_shader_code = include_str!("shaders/vertex.glsl");
+    let vertex_shader_code = include_str!(r#"shaders/vertex.glsl"#);
```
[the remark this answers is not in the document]
Janix4000: It was formatter? I do not remember changing this line...

lib/membrane/video_compositor/implementation/opengl/native/rust.ex
```diff
@@ -63,7 +63,8 @@ end
 
 defmodule Membrane.VideoCompositor.Implementations.OpenGL.Native.Rust.Position do
   @moduledoc """
-  A Position struct describing the video position for use with the rust-based compositor implementation
+  A Position struct describing the video position for use with the rust-based compositor implementation.
+  Position relative to the top right corner of the background video, in the pixels.
```
jerzywilczek, line 67: Maybe this can be rephrased as "Position relative to the top right corner of the viewport, in pixels" or "Position relative to the top right corner of the output, in pixels"

native/membrane_videocompositor_opengl_rust/src/lib.rs
```diff
@@ -50,11 +52,19 @@ mod atoms {
 #[derive(Debug, rustler::NifStruct)]
 #[module = "Membrane.VideoCompositor.Implementations.OpenGL.Native.Rust.RawVideo"]
 struct RawVideo {
-    width: usize,
     height: usize,
+    width: usize,
     pixel_format: rustler::Atom,
 }
 
+/// Position relative to the top right corner of the background video, in the pixels.
```
jerzywilczek, line 60: Same here

native/membrane_videocompositor_opengl_rust/src/lib.rs
```diff
@@ -29,6 +29,8 @@ mod atoms {
       test_atom,
       unsupported_pixel_format,
       I420,
+      function_not_implemented,
+      bad_video_index,
```
jerzywilczek, line 33: I don't think we need this here.

native/membrane_videocompositor_opengl_rust/src/scene.rs
```diff
@@ -211,32 +213,50 @@ impl Scene {
...
+            .insert(video_idx, Video::new(placement, width, height)?);
+        Ok(())
+    }
+
+    /// Remove a video from the scene.
+    pub fn remove_video(&mut self, video_idx: usize) -> Result<(), CompositorError> {
+        self.videos
+            .remove(&video_idx)
+            .ok_or(CompositorError::BadVideoIndex(
+                "video_idx_remove_bad_index",
```
jerzywilczek, line 241: I don't like how this has the word `index` twice.

native/membrane_videocompositor_opengl_rust/src/scene.rs
```diff
@@ -211,32 +213,50 @@ impl Scene {
...
     /// # Panics
     ///
-    /// This function will panic if the `video_idx` is not an index of an existing video and if `data` doesn't have proper length for a YUV420p-encoded frame for the specified video.
+    /// This function will panic if `data` doesn't have proper length for a YUV420p-encoded frame for the specified video.
     pub fn upload_texture(&self, video_idx: usize, data: &[u8]) -> Result<(), CompositorError> {
-        self.videos[video_idx].upload_texture(data)?;
+        self.videos
+            .get(&video_idx)
+            .ok_or(CompositorError::BadVideoIndex(
+                "video_idx_upload_bad_index",
```
jerzywilczek, line 256: I don't like how this has the word `index` twice.

jerzywilczek (changes requested):

native/membrane_videocompositor_opengl_rust/src/lib.rs
jerzywilczek: (I mean `bad_video_index`)

jerzywilczek (approved):
Only looked at rust

pipeline/lib/compose_multiple_inputs.ex
```diff
@@ -0,0 +1,117 @@
...
+    sink
+  end
+
+  defp get_encoder(options) do
+    Map.get(options, :encoder) || NoOp
+  end
+
+  defp get_decoder(options) do
+    Map.get(options, :decoder) || NoOp
+  end
```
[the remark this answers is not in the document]
Janix4000: I changed it in #32

PR #32, opened 2022-09-09
Remove two inputs compitor

Janix4000:
Remove all implementations but `opengl-rust` and adapt tests and benchmarks


lib/membrane/video_compositor/implementations.ex
```diff
@@ -17,7 +17,7 @@ defmodule Membrane.VideoCompositor.MultipleInputs.VideoCompositor.Implementation
         raise ":opengl_cpp is not implemented yet"
```
WojciechBarczynski, line 17: If we are removing ffmpeg, opengl_cpp and nx implementations, shouldn't we remove them here as well?
I think this would be better:
```
case implementation do
  :opengl_rust ->
    {:ok, Membrane.VideoCompositor.OpenGL.Rust}
  :wgpu ->
    {:ok, Membrane.VideoCompositor.Wgpu}
  _other ->
    raise "Format {implemntation} is not supported"
```

DominikWolek (approved):

DominikWolek (approved):

DominikWolek (approved):

PR #33, opened 2022-09-14
A fixup addressing project structure changes.

jerzywilczek:
I didn't notice the files I just merged should have been migrated.


lib/membrane/video_compositor/implementation/wgpu/wgpu.ex
```diff
@@ -0,0 +1 @@
+
```
DominikWolek, line 1: This file is right now unecessary, please delete it

DominikWolek (approved):

PR #35, opened 2022-09-16
Use `Pipeline.Utility.Options` in the benchmarks

Janix4000:
Since the previous PR Pipelines accepts `Options` struct instead of
a raw map. I forgot to update this in the benchmarks code.


WojciechBarczynski (approved):
LGTM

PR #31, opened 2022-09-07
Add an implementation of the `wgpu`-based compositor.

jerzywilczek:

PR #37, opened 2022-09-19
Add dynamic input pads capabilities for the `wgpu`-based compositor.

jerzywilczek:
based on #31

### DIFF
--- a/lib/membrane/video_compositor/implementations/wgpu/native.ex
+++ b/lib/membrane/video_compositor/implementations/wgpu/native.ex
@@ -4,13 +4,30 @@ defmodule Membrane.VideoCompositor.Implementations.Wgpu.Native do
     otp_app: :membrane_video_compositor_plugin,
     crate: "membrane_videocompositor_wgpu"
 
+  alias Membrane.VideoCompositor.Implementations.OpenGL.Native.Rust.Position
   alias Membrane.VideoCompositor.Implementations.OpenGL.Native.Rust.RawVideo
 
-  @spec init(RawVideo.t(), RawVideo.t(), RawVideo.t()) :: {:ok, any()} | {:error, any}
-  def init(_first_video, _second_video, _out_video), do: error()
+  @type internal_state_t :: any()
+  @type error_t :: any()
+  @type id_t() :: non_neg_integer()
 
-  @spec join_frames(any(), binary(), binary()) :: {:ok, binary()} | {:error, any()}
-  def join_frames(_state, _upper, _lower), do: error()
+  @spec init(RawVideo.t()) :: {:ok, internal_state_t()} | {:error, error_t()}
+  def init(_out_video), do: error()
+
+  @spec join_frames(any(), [{id_t(), binary()}]) :: {:ok, binary()} | {:error, error_t()}
+  def join_frames(_state, _input_videos), do: error()
+
+  @spec add_video(internal_state_t(), id_t(), RawVideo.t(), Position.t()) ::
+          :ok | {:error, error_t()}
+  def add_video(_state, _id, _in_video, _position), do: error()
+
+  @spec set_position(internal_state_t(), id_t(), Position.t()) ::
+          :ok | {:error, error_t()}
+  def set_position(_state, _id, _position), do: error()
+
+  @spec remove_video(internal_state_t(), id_t()) ::
+          :ok | {:error, error_t()}
+  def remove_video(_state, _id), do: error()
 
   defp error(), do: :erlang.nif_error(:nif_not_loaded)
 end
--- a/lib/membrane/video_compositor/implementations/wgpu/wgpu.ex
+++ b/lib/membrane/video_compositor/implementations/wgpu/wgpu.ex
@@ -1,0 +1,41 @@
+defmodule Membrane.VideoCompositor.Wgpu do
+  @moduledoc """
+  This module implements video composition in wgpu
+  """
+  @behaviour Membrane.VideoCompositor.FrameCompositor
+
+  alias Membrane.VideoCompositor.Implementations.OpenGL.Native.Rust.Position
+  alias Membrane.VideoCompositor.Implementations.OpenGL.Native.Rust.RawVideo
+  alias Membrane.VideoCompositor.Implementations.Wgpu.Native
+
+  @impl true
+  def init(output_caps) do
+    {:ok, output_caps} = RawVideo.from_membrane_raw_video(output_caps)
+    Native.init(output_caps)
+  end
+
+  @impl true
+  def merge_frames(frames, internal_state) do
+    {Native.join_frames(internal_state, frames), internal_state}
+  end
+
+  @impl true
+  def add_video(id, input_caps, position, internal_state) do
+    {:ok, input_caps} = RawVideo.from_membrane_raw_video(input_caps)
+    {:ok, position} = Position.from_tuple(position)
+    :ok = Native.add_video(internal_state, id, input_caps, position)
+    {:ok, internal_state}
+  end
+
+  @impl true
+  def set_position(id, position, internal_state) do
+    {:ok, position} = Position.from_tuple(position)
+    {Native.set_position(internal_state, id, position), internal_state}
+  end
+
+  @impl true
+  def remove_video(id, internal_state) do
+    :ok = Native.remove_video(internal_state, id)
+    {:ok, internal_state}
+  end
+end
--- a/native/membrane_videocompositor_wgpu/src/compositor.rs
+++ b/native/membrane_videocompositor_wgpu/src/compositor.rs
@@ -1,14 +1,64 @@
+use std::collections::BTreeMap;
+
 mod textures;
 mod videos;
 
 use textures::*;
 use videos::*;
 
+use crate::errors::CompositorError;
+
+/// A point in 2D space
+pub struct Point(pub f32, pub f32);
+
+/// Describes where a video should be located in the scene space.
+/// All coordinates have to be in the range [-1, 1].
+pub struct VideoPlacementTemplate {
+    pub top_right: Point,
+    pub top_left: Point,
+    pub bot_left: Point,
+    pub bot_right: Point,
+    /// This value is supposed to be used for making some videos appear 'in front of' other videos.
+    /// This is still WIP and may not work.
+    pub z_value: f32, // don't really know if setting this will do anything.. I guess it shouldn't without a depth buffer? FIXME??
+}
+
+impl From<VideoPlacementTemplate> for [Vertex; 4] {
+    fn from(template: VideoPlacementTemplate) -> Self {
+        let VideoPlacementTemplate {
+            top_right,
+            top_left,
+            bot_right,
+            bot_left,
+            ..
+        } = template;
+
+        [
+            Vertex {
+                position: [top_right.0, top_right.1, 0.0],
+                texture_coords: [1.0, 0.0],
+            },
+            Vertex {
+                position: [top_left.0, top_left.1, 0.0],
+                texture_coords: [0.0, 0.0],
+            },
+            Vertex {
+                position: [bot_left.0, bot_left.1, 0.0],
+                texture_coords: [0.0, 1.0],
+            },
+            Vertex {
+                position: [bot_right.0, bot_right.1, 0.0],
+                texture_coords: [1.0, 1.0],
+            },
+        ]
+    }
+}
+
 #[repr(C)]
 #[derive(Debug, Clone, Copy, bytemuck::Pod, bytemuck::Zeroable)]
 pub struct Vertex {
-    position: [f32; 3],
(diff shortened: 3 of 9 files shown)


lib/membrane/video_compositor/implementations/wgpu/wgpu.ex
```diff
@@ -0,0 +1,41 @@
...
+  alias Membrane.VideoCompositor.Implementations.Wgpu.Native
+
+  @impl true
+  def init(output_caps) do
+    {:ok, output_caps} = RawVideo.from_membrane_raw_video(output_caps)
+    Native.init(output_caps)
+  end
+
+  @impl true
+  def merge_frames(frames, internal_state) do
```
Janix4000, line 18: Update `internal_state` parameter position. It is the first parameter in almost all callbacks from the `FrameCompositor` behavior. You should rebase from the master.

lib/membrane/video_compositor/implementations/wgpu/wgpu.ex
```diff
@@ -0,0 +1,41 @@
...
+    {Native.join_frames(internal_state, frames), internal_state}
+  end
+
+  @impl true
+  def add_video(id, input_caps, position, internal_state) do
```
Janix4000, line 23: Update `internal_state` parameter position.

lib/membrane/video_compositor/implementations/wgpu/wgpu.ex
```diff
@@ -0,0 +1,41 @@
...
+    {:ok, input_caps} = RawVideo.from_membrane_raw_video(input_caps)
+    {:ok, position} = Position.from_tuple(position)
+    :ok = Native.add_video(internal_state, id, input_caps, position)
+    {:ok, internal_state}
+  end
+
+  @impl true
+  def set_position(id, position, internal_state) do
```
Janix4000, line 31: Update `internal_state` parameter position.

lib/membrane/video_compositor/implementations/wgpu/wgpu.ex
```diff
@@ -0,0 +1,41 @@
...
+    {:ok, position} = Position.from_tuple(position)
+    {Native.set_position(internal_state, id, position), internal_state}
+  end
+
+  @impl true
+  def remove_video(id, internal_state) do
```
Janix4000, line 37: Update `internal_state` parameter position.

Janix4000 (changes requested):

native/membrane_videocompositor_wgpu/src/lib.rs
```diff
@@ -62,41 +58,116 @@ impl InnerState {
...
+        top_right: compositor::Point(right, top),
+        top_left: compositor::Point(left, top),
+        bot_left: compositor::Point(left, bot),
+        bot_right: compositor::Point(right, bot),
+        z_value: 0.0,
+    }
+}
+
+#[rustler::nif]
+fn set_position(
```
WojciechBarczynski, line 149: Can't we just update video.vertices instead of returning error?

native/membrane_videocompositor_wgpu/src/lib.rs
```diff
@@ -10,6 +11,13 @@ pub struct RawVideo {
     pub pixel_format: rustler::Atom,
 }
 
+#[derive(Debug, rustler::NifStruct)]
+#[module = "Membrane.VideoCompositor.Implementations.OpenGL.Native.Rust.Position"]
+struct Position {
```
WojciechBarczynski, line 16: Can't we just use `u32` and convert when needed?

native/membrane_videocompositor_wgpu/src/lib.rs
```diff
@@ -62,41 +58,116 @@ impl InnerState {
...
+}
+
+/// Maps point `x` from the domain \[`x_min`, `x_max`\] to the point in the \[`y_min, y_max`\] line segment, using linear interpolation.
+///
+/// `x` outside the original domain will be extrapolated outside the targe domain.
+fn lerp(x: f64, x_min: f64, x_max: f64, y_min: f64, y_max: f64) -> f64 {
+    (x - x_min) / (x_max - x_min) * (y_max - y_min) + y_min
+}
+
+fn determine_video_placement(
```
WojciechBarczynski, line 114: Maybe this logic should be placed in `native/membrane_videcompositor_wgpu/src/compositor.rs in Scene impl? @DominikWolek

native/membrane_videocompositor_wgpu/src/lib.rs
jerzywilczek: What benefit would that bring?

native/membrane_videocompositor_wgpu/src/lib.rs
jerzywilczek: We could, but it wouldn't be trivial to implement, and this isn't used anywhere, and I think even @Janix4000 agrees that adding this to the `FrameCompositor` behaviour was done a bit too hastily and this isn't very useful or necessary.

native/membrane_videocompositor_wgpu/src/lib.rs
WojciechBarczynski: I think it's just more natural to keep the position as `u32`. I have used it in OpenGL transformations. We need to do type conversion in some place anyway, but in my opinion, doing `u32` to `usize` conversion before indexing arrays improves readability and clearly explains why we need `usize`. When I first looked at the struct with `usize` type in it, I thought it comes from some rustler or WGPU/OpenGL specification.

native/membrane_videocompositor_wgpu/src/lib.rs
WojciechBarczynski: I think that changing video position is one of the key features of a video compositor. As a said on today's daily, I think that such functions should be implemented in `State impl` (maybe with some call to `InputVideo impl`), rather than in `lib.ex` (though I think we need Dominik's opinion on that). We need that feature even if elixir behaviour changes and by implementing it in that way, we'll need to change only integration with elixir.

native/membrane_videocompositor_wgpu/src/compositor.rs
```diff
@@ -1,14 +1,64 @@
...
 mod textures;
 mod videos;
 
 use textures::*;
 use videos::*;
 
+use crate::errors::CompositorError;
+
+/// A point in 2D space
+pub struct Point(pub f32, pub f32);
```
WojciechBarczynski, line 12: Maybe we could add indexing `Point` like: `point.x` `point.y`? I think it looks better and is not that hard to implement. I've done something like this on `video_transforations` branch in OpenGL and calls like `self.placement.top_left.x` in my opinion look better than `self.placement.top_left.0`.
Something like this should work:
```
/// A point in 2D space
pub struct Point {
    pub x: f32,
    pub y: f32
}
```

WojciechBarczynski: BTW can we make it more generic and allow other numerical data types in the `Point` struct such as `u32`. I used `Point` while implementing video transformations in OpenGL and using other types seemed to be a more straightforward solution rather than converting back and forth between `f32` and desired type. [Here is a great example.](https://doc.rust-lang.org/book/ch10-01-syntax.html)

DominikWolek (approved):

native/membrane_videocompositor_wgpu/src/compositor.rs
```diff
@@ -19,60 +69,19 @@ impl Vertex {
...
-        position: [-1.0, -1.0, 0.0],
-        texture_coords: [0.0, 1.0],
-    },
-    Vertex {
-        position: [1.0, -1.0, 0.0],
-        texture_coords: [1.0, 1.0],
-    },
-];
-
 pub struct State {
```
WojciechBarczynski, line 72: I think we should keep something like `output_caps` as `State` attribute or at least scene `width` and `height` (it's useful for some transformations to have information about scene resolution). Maybe we could have two structs `State` and `Scene` and split those responsibilities somehow? I think it's one of the topics to discuss at the next architecture design meeting.

jerzywilczek:
[this](https://github.com/membraneframework-labs/membrane_video_compositor_plugin/pull/31#discussion_r975067848) should be resolved in this PR

native/membrane_videocompositor_wgpu/src/lib.rs
jerzywilczek: I honestly think this is a feature we'll never use in this form and will need to rewrite completely when switching to the new architecture.

lib/membrane/video_compositor/implementations/wgpu/native.ex
```diff
@@ -4,13 +4,30 @@ defmodule Membrane.VideoCompositor.Implementations.Wgpu.Native do
...
+  @type error_t :: any()
+  @type id_t() :: non_neg_integer()
 
-  @spec join_frames(any(), binary(), binary()) :: {:ok, binary()} | {:error, any()}
-  def join_frames(_state, _upper, _lower), do: error()
+  @spec init(RawVideo.t()) :: {:ok, internal_state_t()} | {:error, error_t()}
+  def init(_out_video), do: error()
+
+  @spec join_frames(any(), [{id_t(), binary()}]) :: {:ok, binary()} | {:error, error_t()}
+  def join_frames(_state, _input_videos), do: error()
```
WojciechBarczynski, line 18: Change spec of _state from `any()` to `internal_state_t`. I also sugest changing function name from `join_frames` to something like: `render_frame`, `compose_frames` or something similar.

lib/membrane/video_compositor/implementations/wgpu/wgpu.ex
```diff
@@ -0,0 +1,41 @@
+defmodule Membrane.VideoCompositor.Wgpu do
+  @moduledoc """
+  This module implements video composition in wgpu
+  """
+  @behaviour Membrane.VideoCompositor.FrameCompositor
+
+  alias Membrane.VideoCompositor.Implementations.OpenGL.Native.Rust.Position
+  alias Membrane.VideoCompositor.Implementations.OpenGL.Native.Rust.RawVideo
```
WojciechBarczynski, line 8: 1. You can alias those in one line, using `Membrane.VideoCompositor.Implementations.OpenGL.Native.Rust.{Position, RawVideo}`. 
2. If we are using the same structs for OpenGL and WGPU, we shouldn't name the module `Membrane.VideoCompositor.Implementations.OpenGL.Native.Rust`. Such a name suggests, that this struct is only used in OpenGL implementation. We can separate those structs definitions to other modules like: `Membrane.VideoCompositor.Native.Rust.Position` and `Membrane.VideoCompositor.Native.Rust.RawVideo`.

WojciechBarczynski (changes requested):

native/membrane_videocompositor_wgpu/src/compositor.rs
```diff
@@ -19,60 +38,20 @@ impl Vertex {
...
     device: wgpu::Device,
-    input_videos: [InputVideo; 2],
+    input_videos: BTreeMap<usize, InputVideo>,
     output_textures: OutputTextures,
     pipeline: wgpu::RenderPipeline,
     queue: wgpu::Queue,
     _sampler: wgpu::Sampler,
     sampler_bind_group: wgpu::BindGroup,
+    texture_bind_group_layout: wgpu::BindGroupLayout,
```
WojciechBarczynski, line 49: Can we keep all information about output textures/sampler in dedicated `sampler` and `output_textures` structs? I think keeping `sampler bind group` and `texture bind group layout` in `State` is not the best solution.

WojciechBarczynski (changes requested):
Good job, I've requested some minor changes.

native/membrane_videocompositor_wgpu/src/compositor.rs
jerzywilczek: I'll do that for sampler but not for textures, since textures change a lot in #43 and this can be managed there.

WojciechBarczynski (approved):
GJ, LGTM

lib/membrane/video_compositor/implementations/common.ex
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+  @enforce_keys [:x, :y]
+  defstruct [:x, :y]
+
+  @spec from_tuple({non_neg_integer(), non_neg_integer()}) :: {:ok, __MODULE__.t()}
+  def from_tuple({x, y}) do
+    {:ok, %__MODULE__{x: x, y: y}}
+  end
+end
```
Janix4000, line 49: We have to standardise it. Right now, there are at least two different implementations of the `Position` struct.

Janix4000 (approved):
LGFM

lib/membrane/video_compositor/implementations/common.ex
jerzywilczek: are there?

PR #38, opened 2022-09-19
Fix frame composer internal state parameter order

Janix4000:
Due to an oversight, `FrameComposer` behavior had `internal_state`
as the last argument in all callbacks. We decided this
parameter should always be first.
Additionally, provide better typing into the behavior declarations.


WojciechBarczynski:
LGTM

WojciechBarczynski (approved):

PR #49, opened 2022-10-10
Change the `RGBAToYUVConverter` background colour.

jerzywilczek:
Before, the background colour was green by default (because y = 0, u = 0, v = 0 is green). This patch changes the background colour to black.


Janix4000 (approved):
LGFM

PR #50, opened 2022-10-11
Add z-value sorting when composing videos.

jerzywilczek:
I have to go for now, so this is quite barebones in terms of the API. I do not realy know whether adding z as a field in `Position` is a good solution.

### DIFF
--- a/lib/membrane/video_compositor/frame_compositor.ex
+++ b/lib/membrane/video_compositor/frame_compositor.ex
@@ -22,12 +22,16 @@ defmodule Membrane.VideoCompositor.FrameCompositor do
   @doc """
   Registers a new input video with the given numerical `id`.
   Provided `id` should be unique within all previous ones, otherwise function may cause undefined behaviour.
+  `x` and `y` are pixel coordinates specifying where the top-right corner of the video should be.
+  `z` must be a float between 0.0 and 1.0, and it determines which videos are drawn in front of others.
+  A video with a lower `z` coordinate will cover videos with higher `z` coordinates.
   """
   @callback add_video(
               internal_state :: internal_state_t,
               id :: id_t(),
               input_caps :: RawVideo.t(),
-              position :: {x :: non_neg_integer(), y :: non_neg_integer(), z :: float()}
+              position :: {x :: non_neg_integer(), y :: non_neg_integer()},
+              z :: float()
             ) :: {:ok, internal_state_t} | {:error, error_t()}
 
   @doc """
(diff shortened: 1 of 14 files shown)


lib/membrane/video_compositor/implementations/common.ex
```diff
@@ -36,14 +36,15 @@ defmodule Membrane.VideoCompositor.Implementations.Common.Position do
...
-  defstruct [:x, :y]
+  @enforce_keys [:x, :y, :z]
+  defstruct [:x, :y, :z]
 
-  @spec from_tuple({non_neg_integer(), non_neg_integer()}) :: {:ok, __MODULE__.t()}
-  def from_tuple({x, y}) do
-    {:ok, %__MODULE__{x: x, y: y}}
+  @spec from_tuple({non_neg_integer(), non_neg_integer(), float()}) :: {:ok, __MODULE__.t()}
+  def from_tuple({x, y, z}) do
+    {:ok, %__MODULE__{x: x, y: y, z: z}}
```
DominikWolek, line 48: Shouldn't we check z here if it is in <-1,1> range?

lib/membrane/video_compositor/frame_compositor.ex
```diff
@@ -27,7 +27,7 @@ defmodule Membrane.VideoCompositor.FrameCompositor do
               internal_state :: internal_state_t,
               id :: id_t(),
               input_caps :: RawVideo.t(),
-              position :: {x :: non_neg_integer(), y :: non_neg_integer()}
+              position :: {x :: non_neg_integer(), y :: non_neg_integer(), z :: float()}
```
Janix4000, line 30: 1. In many 2D frameworks z-index is an integer
3. At Elixir's level of abstraction 'third dimension' has no sense, it should be a separate field

Janix4000 (changes requested):

Janix4000 (changes requested):
Minor changes and z ordering

lib/membrane/video_compositor/implementations/common.ex
```diff
@@ -46,11 +46,11 @@ defmodule Membrane.VideoCompositor.Implementations.Common.Position do
...
+    {:error, :z_out_of_range}
+  end
+
   def from_tuple({x, y, z}) do
-    if z < 0.0 or z > 1.0 do
-      {:error, :z_out_of_range}
-    else
-      {:ok, %__MODULE__{x: x, y: y, z: z}}
-    end
+    {:ok, %__MODULE__{x: x, y: y, z: z}}
```
Janix4000, line 54: Maybe we should raise exceptions, instead of the `{:ok | :error, ...}` pattern, @DominikWolek what do you think?

native/membrane_videocompositor_wgpu/src/lib.rs
```diff
@@ -108,7 +108,13 @@ fn add_video(
...
             width: input_video.width,
             height: input_video.height,
-            z: position.z,
+            // we need to do this because 0.0 is an intuitively standard value and maps onto 1.0,
+            // which is outside of the wgpu clip space
+            z: if position.z == 0.0 {
+                1.0 - 1e-7
+            } else {
+                1.0 - position.z
+            },
```
Janix4000, line 117:
```suggestion
            z: 1.0 - max(position.z, 1e-7)
```

Janix4000 (changes requested):
Small fixes

Janix4000 (approved):

PR #60, opened 2022-11-14
Ms 274 live queueing video frames

jerzywilczek:
This patch makes use of Membrane timers to allow live operation of the
compositor:

 * A timer with interval equal to seconds / frames is started when the element starts operating.

 * Whenever a frame is composed naturally (when all vidoes have frames ready for composition) the timer is restarted.

 * When the timer ticks, a frame is forcibly rendered, using previous frames as fallback in cases where we don't have a frame ready.

 * To protect against stale frames clogging the front of the queue, whenever a frame is rendered stale frames are removed from the front of the queue.
 
 It also adds tests:
 
 *  A new element, `BadConnectionEmulator`, was added to simulate delays and
packet loss.

 * The `ComposeMultipleInputs` pipeline was altered to allow introducing an
additional filter between the decoder and the compositor.

 * A new test was added (`packet_loss.ex`). It is completely analogous to
the h264 pipeline test, but uses the `BadConnectionEmulator` and sets
the compositor to live mode.

These tests don't verify much unfortunately. Since some frames are
randomly dropped or delayed, there is no reference result to compare to.


lib/membrane/compositor_element.ex
```diff
@@ -6,12 +6,18 @@ defmodule Membrane.VideoCompositor.CompositorElement do
...
   def_options caps: [
-                type: RawVideo,
+                spec: RawVideo.t(),
                 description: "Struct with video width, height, framerate and pixel format."
+              ],
+              live: [
+                spec: boolean(),
+                description: "Set the compositor to live mode",
+                default: false
               ]
```
Janix4000, line 21: Does `live` option trigger frames dropping? I think it should be better described.

Janix4000 (changes requested):

lib/membrane/compositor_element.ex
```diff
@@ -143,7 +187,11 @@ defmodule Membrane.VideoCompositor.CompositorElement do
     state = %{state | internal_state: internal_state}
 
     if all_input_pads_received_end_of_stream?(context.pads) do
-      {{:ok, end_of_stream: :output}, state}
+      if state.live do
+        {{:ok, end_of_stream: :output, stop_timer: :render_frame}, state}
+      else
+        {{:ok, end_of_stream: :output}, state}
+      end
```
DominikWolek, line 194:
```suggestion
     actions = if all_input_pads_received_end_of_stream?(context.pads) do
      stop = if state.live, do: [stop_timer: :render_frame], else: []
      [end_of_stream: :output] ++ stop
    else
      []
    end

    {{:ok, actions}, state}
```
Obviously it is also for the line above and below, but GitHub don't let do this proper way.

DominikWolek (changes requested):

lib/membrane/compositor_element.ex
```diff
@@ -1,17 +1,34 @@
...
+
+  It is capable of operating in one of two modes:
+
+   * offline compositing:
+     The compositor will wait for all videos to have a recent enough frame available and then perform the compositing.
+
+   * live compositing:
+     In this mode, if the compositor will start a timer ticking every spf (seconds per frame). The timer is reset every time a frame is produced.
+     If the compositor doesn't have all frames ready by the time the timer ticks, it will produce a frame anyway, using old frames as fallback in cases when a current frame is not available.
+     If the frames arrive later, they will be dropped.
```
DominikWolek, line 13: Maybe mention updating the fallback frame

lib/membrane/compositor_element.ex
```diff
@@ -1,17 +1,34 @@
...
+  alias Membrane.Buffer
   alias Membrane.RawVideo
   alias Membrane.VideoCompositor.Wgpu
 
   def_options caps: [
-                type: RawVideo,
+                spec: RawVideo.t(),
                 description: "Struct with video width, height, framerate and pixel format."
+              ],
+              live: [
```
DominikWolek, line 26: I know we kept calling this feature "live queueing" but I wonder if real-time isn't a better name for this.

DominikWolek (approved):

Janix4000 (approved):
Good job

PR #61, opened 2022-11-21
Remove information about compiling Google ANGLE from the readme.

jerzywilczek:
This is unnecessary since we don't use OpenGL anymore.


Janix4000 (approved):
Finally

PR #64, opened 2022-11-24
Add ptf offset option for input stream

WojciechBarczynski:


lib/membrane/compositor_element.ex
```diff
@@ -43,6 +43,11 @@ defmodule Membrane.VideoCompositor.CompositorElement do
         description:
           "Initial position of the video on the screen, given in the pixels, relative to the upper left corner of the screen",
         default: {0, 0}
+      ],
+      pts_offset: [
+        spec: Membrane.Time.t(),
+        description: "Input stream PTS offset in nanoseconds",
+        default: 0
```
DominikWolek, line 50: I think this should be renamed to `offset` - similarly how this option is named in the `audio_mix_plugin`. This applies to names of variables and fields, too.

DominikWolek (changes requested):

lib/membrane/compositor_element.ex
```diff
@@ -43,6 +43,11 @@ defmodule Membrane.VideoCompositor.CompositorElement do
         description:
           "Initial position of the video on the screen, given in the pixels, relative to the upper left corner of the screen",
         default: {0, 0}
+      ],
+      pts_offset: [
+        spec: Membrane.Time.t(),
```
DominikWolek, line 48:
```suggestion
        spec: Membrane.Time.non_neg_t(),
```

lib/membrane/compositor_element.ex
```diff
@@ -43,6 +43,11 @@ defmodule Membrane.VideoCompositor.CompositorElement do
...
+        description: "Input stream PTS offset in nanoseconds",
```
DominikWolek, line 49:
```suggestion
        description: "Input stream PTS offset in nanoseconds. Must be non-negative.",
```

DominikWolek (changes requested):

lib/membrane/compositor_element.ex
WojciechBarczynski: I've changed it, but I think that `pts_offset` is a more descriptive and less confusing name for it. In `audio_mix_plugin` we don't encounter other types of offsets, but here at first glance, it might be interpreted another way, since the `position` option refers to final video placement.

lib/membrane/compositor_element.ex
WojciechBarczynski: Done

lib/membrane/compositor_element.ex
WojciechBarczynski: Done

lib/membrane/compositor_element.ex
DominikWolek: Ok, I agree. `timestamp_offset` is fine, then?

WojciechBarczynski: Sounds great :)

WojciechBarczynski: Done

lib/membrane/compositor_element.ex
```diff
@@ -93,17 +99,19 @@ defmodule Membrane.VideoCompositor.CompositorElement do
   @impl true
   def handle_pad_added(pad, context, state) do
     position = context.options.position
+    timestamp_offset = context.options.timestamp_offset
```
DominikWolek, line 102: Does it work? Shouldn't that be rather `context.pads.pad.options.timestamp_offset`?

DominikWolek (changes requested):

lib/membrane/compositor_element.ex
WojciechBarczynski: It does. Here is how Membrane.Element.CallbackContext.PadAdded struct looks like in `handle_pad_added`:
```elixir
%Membrane.Element.CallbackContext.PadAdded{
  direction: :input,
  options: %{position: {0, 0}, timestamp_offset: 2},
  pads: %{
    {Membrane.Pad, :input, #Reference<0.726837243.177209346.24902>} => %Membrane.Element.PadData{
      accepted_caps: {Membrane.RawVideo, [pixel_format: :I420]},
      availability: :on_request,
      caps: nil,
      start_of_stream?: false,
      end_of_stream?: false,
      direction: :input,
      mode: :pull,
      name: :input,
      ref: {Membrane.Pad, :input, #Reference<0.726837243.177209346.24902>},
      options: %{position: {0, 0}, timestamp_offset: 2},
      pid: #PID<0.725.0>,
      other_ref: :output,
      input_queue: nil,
      demand: 4000,
      demand_mode: :auto,
      demand_unit: :buffers,
      other_demand_unit: :buffers,
      auto_demand_size: 4000,
      sticky_messages: [],
      toilet: nil,
      associated_pads: []
    }
  },
  playback_state: :stopped,
  clock: nil,
  parent_clock: #PID<0.717.0>,
  name: :compositor
}
```

DominikWolek (approved):

PR #80, opened 2023-01-16
Make the corner rounding shader use anti-aliasing and be branchless.

jerzywilczek:
This patch removes all if statements from the corner rounding shader in favor of arithmetic operations (which is optimal for shaders).

The shader also uses a very basic form of anti-aliasing now. Basically it applies the `smoothstep` function to the alpha value of pixels on the corners, which "fades them" gently instead of cutting them off sharply.

### DIFF
--- a/native/membrane_videocompositor/src/compositor/texture_transformations/corners_rounding/corners_rounding.wgsl
+++ b/native/membrane_videocompositor/src/compositor/texture_transformations/corners_rounding/corners_rounding.wgsl
@@ -34,10 +34,10 @@ var sampler_: sampler;
 var<uniform> corners_rounding_uniform: CornersRoundingUnifrom;
 
 struct IsInCorner {
-    left_border: bool,
-    right_border: bool,
-    top_border: bool,
-    bot_border: bool,
+    left_border: f32,
+    right_border: f32,
+    top_border: f32,
+    bot_border: f32,
 }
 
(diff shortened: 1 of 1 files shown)


native/membrane_videocompositor/src/compositor/texture_transformations/corners_rounding/corners_rounding.wgsl
```diff
@@ -46,52 +46,44 @@ fn get_nearest_inner_corner_coords_in_pixels(
...
+        border_radius
+    );
+
+    let d = distance(input.texture_coords * vec2(video_width, video_height), corner_coords);
+
+    let anti_aliasing_pixels = 1.5;
+
+    let alpha = smoothstep(border_radius + anti_aliasing_pixels, border_radius - anti_aliasing_pixels, d);
+
+    return vec4(colour.xyz, is_in_corner * alpha + (1. - is_in_corner));
```
WojciechBarczynski, line 88: Since you are performing AA here, we should change ref video in texture transformation test. I've updated ref for mac, but you have to update linux one as well (I know that diff i below error margin, but we should have accurate video on repo)

WojciechBarczynski (changes requested):
Update ref. The rest is ok

WojciechBarczynski (approved):
gj 💪🏻

PR #82, opened 2023-01-20
Release fix

mat-hek:


mix.exs
```diff
@@ -78,7 +78,13 @@ defmodule Membrane.VideoCompositor.Mixfile do
       links: %{
```
DominikWolek, line 78: I'm little bit confused: I thought version 0.1 was supposed to be based on Core 0.10. Are our users (notably the Live team) ok with Core 0.11?

mix.exs
```diff
@@ -88,7 +94,15 @@ defmodule Membrane.VideoCompositor.Mixfile do
...
-      nest_modules_by_prefix: [Membrane.VideoCompositor]
+      nest_modules_by_prefix: [
+        Membrane.VideoCompositor,
+        Membrane.VideoCompositor.VideoTransformations
+      ],
+      groups_for_modules: [
+        "Video transformations": [
+          ~r/^Membrane\.VideoCompositor\.VideoTransformations($|\.)/
+        ]
+      ]
```
DominikWolek, line 105: It does not exatly fit into side bar  
image

DominikWolek (changes requested):

mix.exs
mat-hek: AFAIK they needed a released compositor compatible with core 0.11, what seems reasonable since they're currently moving to 0.11 ;)

mix.exs
mat-hek: That's true, however, I think it's not possible to fit them - at least either the header (`Membrane.VideoCompositor.VideoTr...`) is cut or the module names are. I think the solution would be to shorten these java-like names, for example by removing the `.VideoTransformations` part, though it's out of scope of this PR.

README.md
```diff
@@ -4,6 +4,8 @@
 [![API Docs](https://img.shields.io/badge/api-docs-yellow.svg?style=flat)](https://hexdocs.pm/membrane_video_compositor_plugin) -->
 [![CircleCI](https://dl.circleci.com/status-badge/img/gh/membraneframework-labs/membrane_video_compositor_plugin/tree/master.svg?style=svg)](https://dl.circleci.com/status-badge/redirect/gh/membraneframework-labs/membrane_video_compositor_plugin/tree/master)
 
+Membrane plugin that accepts multiple video inputs, transforms them according to provided transformations and composes them into a single output video.
+
 It is part of [Membrane Multimedia Framework](https://membraneframework.org).
```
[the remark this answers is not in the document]
DominikWolek: I guess you have to write it from he scratch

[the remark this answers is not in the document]
mat-hek: Hmm, so we can take it from tests, but they're quite sophisticated and I created a story to simplify them: https://membraneframework.atlassian.net/browse/MS-380 So maybe we could solve that first and then create an example based on that?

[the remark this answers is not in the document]
DominikWolek: Alright then

DominikWolek (approved):

PR #83, opened 2023-01-23
Scene

DominikWolek:
Here is some key API info:

- All VC processing is defined by the Scene struct - One overly complicated set of random things to rule them all
- There are three types of Objects (renderable things): InputVideos, Textures and Layouts
- InputVideos wrap input pads as a processing graph node
- Texture objects are single input and single output nodes in a processing graph
- Layout objects are multiple inputs and a single output nodes in a processing graph
- Objects take other Objects as inputs
- All Objects can be imagined as things taking the simple image(s), processing them, and returning new image
- User of VC can create custom Transformations (applied to textures) and Layouts (hopefully writing only a little of Elixir, Rust, and shader code


jerzywilczek (changes requested):
I don't get most of this. What is the difference between `Manipulation`, `Alternation`, `Transformation`, `Compound` and `Layout`? why is there so many of these? the names also are not self-explanatory and without documentation I have no way of telling which one does what.

test/example/scene_example.ex
```diff
@@ -0,0 +1,73 @@
...
+        inputs_map: %{
+          rotated: :top_left,
+          rounded: :top_right,
+          ball: :bottom
+        }
+      }
+    ],
+    render: :final_object
+  }
+end
```
WojciechBarczynski, line 73: It isn't clear to me what the output video should look like, even though I've opened the jamboard from pair programming next to it :). Especially `Compound` structs are confusing. Simple one/two-lined comments describing each example would be helpful. I've more or less understood what is happening here after 15min of analyzing, which IMO is not great for an API example :)

lib/membrane/video_compositor/compound.ex
```diff
@@ -0,0 +1,22 @@
...
+  @enforce_keys [:base_size, :inputs_map, :layout]
+  defstruct @enforce_keys
+
+  @type t :: %__MODULE__{
+          # base_size: BaseSize.t() | Object.input_t(),
+          base_size: BaseSize.t(),
+          inputs_map: %{Object.name_t() => Layout.input_t()},
+          layout: Layout.name_t()
+        }
+end
```
WojciechBarczynski, line 22: What `Compounds` can take as input? Maybe we should have some way of specifying whether inputs should be `Texture`, `Canvas`, or other `Compounds` types? Otherwise, I can't think of a simple way of specifying what type of input is accepted by certain `Compound`.

WojciechBarczynski (changes requested):
Overall I like this API, even though I think it doesn't necessarily follow the `make simple things easy` motto :)

test/example/scene_example.ex
```diff
@@ -0,0 +1,73 @@
...
+          rounded_3: :bottom
+        },
+        layout: :strip
+      },
+      final_object: %Compound{
+        base_size: %BaseSize{height: 1080, width: 1920},
+        inputs_map: %{
+          1.0 => :strip,
+          0.0 => :rounded_4
+        },
```
WojciechBarczynski, line 35: 1.0, 0.0? wtf?

WojciechBarczynski (changes requested):

test/example/scene_example.ex
WojciechBarczynski: Using floats as keys in a map is a pretty bad idea, but here it's even worst since it's hard to figure out what that means.

DominikWolek: 1.0 and 0.0 are simply z-values, the strip simply is put on top of rounded 4th video. But I can change that example.

lib/membrane/video_compositor/compound.ex
DominikWolek: The idea was to always render objects as a texture after all transformations adn pass it to the next objects. Compounds de facto will always take a texture as an input.

lib/membrane/video_compositor/compound/layout.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Membrane.VideoCompositor.Compound.Layout do
+  @moduledoc """
+  Behaviour representing layous - modification combing multiple objects into one.
+
+  Render callback is a placeholder.
+  """
+
+  @type definition_t :: struct() | module()
```
WojciechBarczynski, line 8: I thought that in Elixir we usually use:
```suggestion
  @type t :: struct() | module()
```

lib/membrane/video_compositor/scene.ex
```diff
@@ -0,0 +1,26 @@
...
+  alias Membrane.VideoCompositor.Object
+  alias Membrane.VideoCompositor.Object.SimpleAlternation
+
+  @enforce_keys [:objects, :render]
+  defstruct @enforce_keys ++ [alternations: [], layouts: []]
+
+  @type t :: %__MODULE__{
+          alternations: [{SimpleAlternation.name_t(), SimpleAlternation.definition_t()}],
+          layouts: [{Layout.name_t(), Layout.definition_t()}],
+          objects: [{Object.name_t(), Object.t()}],
```
WojciechBarczynski, line 17: For consistency:
```suggestion
          layouts: [{Layout.name_t(), Layout.t()}],
          objects: [{Object.name_t(), Object.t()}],
```

WojciechBarczynski (changes requested):

lib/membrane/video_compositor/compound/layout.ex
```diff
@@ -0,0 +1,13 @@
...
+  @type name_t :: any()
+  @type input_t :: any()
```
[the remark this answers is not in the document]
DominikWolek: - `input_t` should be `any` to allow to define custom input maps (e.g. floats for z-values, ints for some kind of order, etc.). Maybe the type should be renamed to not be confused with `Object.input_t`?
-  I guess we could limit `name_t` types to tuple or atom, similarly to `name_t` types of Membrane Bins, Elements, etc.

lib/membrane/video_compositor/scene.ex
DominikWolek: As above

lib/membrane/video_compositor/compound/layout.ex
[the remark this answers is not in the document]
DominikWolek: I don't see a reason to allow `String.t()` or `Pad.ref()` as name, though

test/example/scene_example.ex
```diff
@@ -0,0 +1,87 @@
...
+          rounded_3: :bottom
+        },
+        layout: :strip
+      },
+      final_object: %Compound{
+        base_size: %BaseSize{height: 1080, width: 1920},
+        inputs_map: %{
+          strip: 1.0,
+          rounded_4: 0.0
+        },
```
jerzywilczek, line 49: I'm a bit unsure about this `inputs_map` thing, but I don't have any better ideas.

jerzywilczek (approved):
to me, it seems mostly good, but I'm rather bad at API design. Other ppl, like @varsill and @Janix4000 should take a look.

lib/membrane/video_compositor/compound/layout.ex
WojciechBarczynski: ok

lib/membrane/video_compositor/compound/layout.ex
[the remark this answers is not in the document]
WojciechBarczynski: Ok

WojciechBarczynski (approved):

test/example/scene_example.ex
```diff
@@ -0,0 +1,87 @@
...
+      rounded_2: %Texture{input: :video_2, transformations: [:rounding]},
+      rounded_3: %Texture{input: :video_3, transformations: [:rounding]},
+      rounded_4: %Texture{input: :video_4, transformations: [:rounding]},
+      strip: %Compound{
+        base_size: %BaseSize{height: 1080, width: 480},
+        inputs_map: %{
+          rounded_1: :top,
+          rounded_2: :middle,
+          rounded_3: :bottom
+        },
```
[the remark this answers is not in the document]
Janix4000: Why videos are supposed to be scaled accordingly to the size of the container?
I can easily imagine the left scenario instead of the right one.
image

Janix4000 (commented):
I'm unsure about this design

test/example/scene_example.ex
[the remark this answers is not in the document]
DominikWolek: Z-values of `strip` and `rounded`

test/example/scene_example.ex
```diff
@@ -0,0 +1,87 @@
...
+    ],
+    layouts: [
+      overlay: Overlay,
+      strip: Grid
+    ],
+    objects: [
+      rounded_1: %Texture{input: :video_1, transformations: [:rounding]},
+      rounded_2: %Texture{input: :video_2, transformations: [:rounding]},
+      rounded_3: %Texture{input: :video_3, transformations: [:rounding]},
+      rounded_4: %Texture{input: :video_4, transformations: [:rounding]},
```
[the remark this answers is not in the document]
DominikWolek: is this a comment on the example or type specification?

If the latter: I don't see why restrict users to ONLY tuples. `name_t :: tuple() | atom()` allow user to choose between simplicity of atoms and manageability of tuples.

test/example/scene_example.ex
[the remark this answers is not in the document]
DominikWolek: - `Compound` is an `Object` with multiple inputs
- `Layout` is behaviour of `Compound` modifications
- `Grid` is specific `Layout`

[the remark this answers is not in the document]
DominikWolek: > Why videos are supposed to be scaled accordingly to the size of the container?
I can easily imagine the left scenario instead of the right one.

Of course, scenerio on the right is also possible. Scenerio on the right is just, imho, simpler and more suited for the example

test/example/scene_example.ex
```diff
@@ -0,0 +1,87 @@
...
+        base_size: %BaseSize{height: 1080, width: 1920},
+        inputs_map: %{
+          strip: 1.0,
+          rounded_4: 0.0
+        },
+        layout: :overlay
+      }
+    ],
+    render: :final_object
+  }
```
bblaszkow06, line 54: As you can see, `inputs_map` generates quite confusion. I wonder what is the reason for Compound to exist in the first place? IMO layouts would be enough and they could provide better APIs for defining inputs, e.g.

```elixir
%Overlay{
  base_size: %BaseSize{height: 1080, width: 1920},
  layers: [
    # position in lists determines the order of render, no need to add z-index
    # 2nd and 3rd item are x & y positions of input
    {:input1, 0, 0},
    {:input2, 0, 0}
  ]
}
%AdaptiveGrid{
  base_size: %BaseSize{height: 1080, width: 1920},
  inputs: [:input1, :input2, :input3, :input4]
}
# Grid with one main input and the rest grouped on the side
%PinnedItemGrid{
  base_size: %BaseSize{height: 1080, width: 1920},
  main: :input1,
  side: [:input2, :input3, :input4]
}
```

lib/membrane/video_compositor/canvas.ex
```diff
@@ -0,0 +1,16 @@
+defmodule Membrane.VideoCompositor.Canvas do
+  @moduledoc """
+  Structure representing Canvas objects.
```
bblaszkow06, line 3: But what is Canvas object?

lib/membrane/video_compositor/compound/base_size.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Membrane.VideoCompositor.Compound.BaseSize do
```
bblaszkow06, line 1: 1. Why `BaseSize`? I mean - what the `Base` refers to here?
2. Isn't it generic enough to define it at the compositor's top level?

bblaszkow06 (changes requested):

lib/membrane/video_compositor/canvas/transformation.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Membrane.VideoCompositor.Canvas.Transformation do
+  @moduledoc """
+  Behaviour representing alternations of a Canvas object.
+  Under the hood canvas transformation most likely are implemented as a vertex shaders.
```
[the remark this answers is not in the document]
jerzywilczek:
```suggestion
  Under the hood canvas transformations are most likely implemented as a vertex shader.
```

WojciechBarczynski:
Looking back at this API, I think we can simplify it a little bit. From the user's point of view, I hardly see any major benefits of separating the `canvas` and `texture` objects, when `canvas` objects require specifying `input`. Maybe we can just have an `object` with `transformation` and `compound` (or `layout`) taking multiple `objects` and returning the single one. We can deal with differences in transformations under the hood. I think that such API would better match the `make simple things easy and hard things possible` motto.

WojciechBarczynski (changes requested):
After thinking about this API and its low-level implementation for a longer time, I think this API can and should be simplified. I don't have any reasonable idea of how to implement canvases to be intuitive to use, moreover in some cases, I'm not sure whether to use canvas or texture objects. Maybe we can talk about some ideas on how to change this API, that we came up with @jerzywilczek?

lib/membrane/video_compositor/canvas/transformation.ex
varsill, line 4:
```suggestion
  Under the hood canvas transformations most likely are implemented with the use of vertex shaders.
```

lib/membrane/video_compositor/canvas/transformation.ex
```diff
@@ -0,0 +1,13 @@
...
+
+  Render callback is a placeholder.
```
varsill, line 6: I don't think I get what you mean there

lib/membrane/video_compositor/compound/layout.ex
varsill, line 8: Are you sure we want to stick to the convention of types having `_t` suffix? Recently we have decided to get rid of all the `_t` suffixes in Membrane Core as we have realised nobody is following such a convention - but I don't know if there is any "Membrane-wide" agreement on that :D

test/example/scene_example.ex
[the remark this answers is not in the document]
varsill: IMHO, referring to `compound` object's behavior as `layout` is confusing - I think we should either remove `layout` term from the API, or the `compound` term

lib/membrane/video_compositor/compound/layout.ex
DominikWolek: I like type names without `_t` - let's talk about Membrane-wide convention on the next brainstorming

lib/membrane/video_compositor/examples/mock/layouts/overlay.ex
```diff
@@ -0,0 +1,24 @@
...
+  @moduledoc false
+
+  alias Membrane.VideoCompositor.Examples.Mock.Layouts.Position
+  alias Membrane.VideoCompositor.Scene.Object
+
+  @enforce_keys [:overlay_spec, :inputs, :resolution]
+  defstruct @enforce_keys
+
+  @typedoc """
+  Name used to identify create placing name - position relationship.
```
varsill, line 11: I don't think I do understand :<

lib/membrane/video_compositor/scene/layout.ex
```diff
@@ -0,0 +1,42 @@
+defmodule Membrane.VideoCompositor.Scene.Layout do
+  @moduledoc """
+  Structure representing Layout objects.
+
+  Layouts can take multiple renderable objects as input and
+  combine them into one frame. Fadings, Grids, Overlays, Transitions,
+  etc. can be defined as Layouts.
+
+  Basically multi-input, single-output node in processing graph.
```
varsill, line 9:
```suggestion
  Basically it's a multi-input, single-output node in processing graph.
```

lib/membrane/video_compositor/scene/layout.ex
```diff
@@ -0,0 +1,42 @@
...
+  """
+  alias Membrane.VideoCompositor.Scene.Object
+
+  @typedoc """
+  Keys enforced in Layout objects.
+  """
+  @type enforced_keys :: :inputs | :resolution
+
+  @typedoc """
+  Specify :inputs types in Layout structs.
```
varsill, line 19: I think it would be better to focus on semantics of that type rather than the fact that it's used in the `:inputs` map.

lib/membrane/video_compositor/scene/object.ex
```diff
@@ -0,0 +1,38 @@
...
+  @type name :: tuple() | atom()
+
+  @typedoc """
+  Defines how the input of an object can be specified
+  in Video Compositor.
+  """
+  @type input :: name() | Pad.name_t()
+
+  @typedoc """
+  Define how the output resolution of an object can be specified.
```
varsill, line 31:
```suggestion
  Defines how the output resolution of an object can be specified.
```

lib/membrane/video_compositor/scene/object.ex
```diff
@@ -0,0 +1,38 @@
...
+  Additionally, in Textures resolution can be specified as
```
varsill, line 32: How about defining an output_resolution for Textures and a separate output_resolution for Layouts, and simply defining `Object.object_output_resolution` as an alternative of these two types?
I believe it would help in separating the "types of resolutions" applicable to different object - the information that "...the Textures resolution can be specified as as a transformed resolution of the object" could be then available in the Texture module documentation.

lib/membrane/video_compositor/scene/texture.ex
```diff
@@ -0,0 +1,24 @@
+defmodule Membrane.VideoCompositor.Scene.Texture do
+  @moduledoc """
+  Texture take frame received from Video Compositor objects,
+  apply all transformations and can be passed as an input
+  for other objects.
```
varsill, line 5:
```suggestion
  Texture takes a frame received from Video Compositor objects,
  applies all transformations and can be passed as an input
  to other objects.
```

lib/membrane/video_compositor/scene/texture.ex
```diff
@@ -0,0 +1,24 @@
+defmodule Membrane.VideoCompositor.Scene.Texture do
```
varsill, line 1: What is the genesis of a `Texture` name? :D 
To be honest, it's not intuitive for me to call it that way - as far as I know, a Texture is just a sequence of Transformations (with an ability to set a "fixed" output resolution).

lib/membrane/video_compositor/scene/transformation.ex
```diff
@@ -0,0 +1,19 @@
...
+  @typedoc """
+  Specify how Textures can be defined:
+    - By struct - when transformation can be parametrized with different values
+    e.g. corners round (parametrized with border-radius), cropping, etc.
+    - By module - when there are no reasonable / common use cases of parametrization,
+    and identifying transformation without it is enough e.g. RollToBall - transformation
+    turning input object / frame / video into the ball
+  """
+  @type definition :: struct() | module()
+  @type name :: tuple() | atom()
```
varsill, line 18: Shouldn't you refer to the `Object.name()` here?

lib/membrane/video_compositor/scene/transformation.ex
```diff
@@ -0,0 +1,19 @@
+defmodule Membrane.VideoCompositor.Scene.Transformation do
+  @moduledoc """
+  Behaviour representing single input, single output transformations of objects.
```
varsill, line 3:
```suggestion
  Behaviour representing single input, single output transformation of objects.
```

varsill: And are you sure we can call it a "behaviour"? In Elixir term "behaviour" has a very strict meaning.

lib/membrane/video_compositor/scene/texture.ex
DominikWolek: I guess as some sort of mirror to `Canvas` object: both have set of transformations and use single input, but one operates on canvases, the other - on textures. 

Name might be a little redundant for now, but I guess it describes what's under the hood. You've got your texture on GPU from `:input`, you transform it according to the `:transformations` and you want the output texture to have resolution specified under the `:resolution`.

lib/membrane/video_compositor/scene/transformation.ex
DominikWolek: Well, as Tranformation is not an object, no.

lib/membrane/video_compositor/scene/transformation.ex
DominikWolek: As we aren't sure if Transformation will end up as a behaviour, i'll change it to the module

lib/membrane/video_compositor/examples/hard.ex
```diff
@@ -0,0 +1,50 @@
...
+  %Scene{
+    objects: [
+      rotated: %Texture{
+        input: :video_1,
+        transformations: [@rotate]
+      },
+      merged: %Merging{
+        inputs: %{
+          first: :video_1,
+          second: :video_2
```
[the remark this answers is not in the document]
DominikWolek: No, as to show you can use same video (or object) multiple times.

pipeline/lib/membrane/video_compositor/pipeline/utility/options.ex
```diff
@@ -6,46 +6,46 @@ defmodule Membrane.VideoCompositor.Pipeline.Utils.Options do
...
-          decoder: decoder_t(),
-          encoder: encoder_t(),
-          input_filter: input_filter_t()
+          inputs: inputs(),
+          output: output(),
+          stream_format: stream_format(),
+          compositor: compositor(),
+          decoder: decoder(),
+          encoder: encoder(),
+          input_filter: input_filter()
```
WojciechBarczynski, line 48: All those `something_t` -> `something` type name changes IMO are not relevant and outside of the scope of this PR. Those changes have nothing to do with scene API design.

lib/membrane/video_compositor/scene/texture.ex
WojciechBarczynski: As @DominikWolek explained, the name describes what is happening under the hood. In my first design, I used `Transformations` and `Layouts` as processing graph nodes, which took `Textures` as inputs. Since many times we chain multiple single-input single-output augmentations  (like cropping and corner rounding), it's more compact to have a node that can perform any number of such operations, than having to define each of them separately.

It's also why `Transformations` are not `Objects`.  We wrap them into `Textures` in `Scene` `objects`.

I think that the `Texture` name clarifies why we have to specify output `resolution`.

lib/membrane/video_compositor/examples/hard.ex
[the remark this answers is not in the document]
WojciechBarczynski:
``` elixir
defmodule Membrane.VideoCompositor.Examples.Hard do
  @moduledoc """
  ...
  - The second video is merged with the first video. The result gets
  corner rounding and is put in the top right corner
  ...
  """```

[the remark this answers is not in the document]
Comment (varsill) [reply lib/membrane/video_compositor/examples/hard.ex:33]: Oh, I got it now ;)

**Review by varsill — COMMENTED** [review]
WDYT about moving Texture and Layout modules into Membrane.VideoCompositor.Scene.Object namespace? At the first glance it's not clear to me now what is the relationship between Object and Texture/Layout.
I would also consider adding a general description of these entities and relationships between them in the `scene.ex` file

**Comment by DominikWolek** [comment]
> WDYT about moving Texture and Layout modules into Membrane.VideoCompositor.Scene.Object namespace? At the first glance it's not clear to me now what is the relationship between Object and Texture/Layout. I would also consider adding a general description of these entities and relationships between them in the `scene.ex` file

That's not a bad idea

File: lib/membrane/video_compositor/examples/easy.ex [hunk 26]
```diff
@@ -0,0 +1,81 @@
...
+      }
+    },
+    inputs: %{},
+    resolution: nil
+  }
+
+  %Scene{
+    objects: [
+      rounded_1: %Texture{
+        input: :video_1,
```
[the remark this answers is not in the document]
WojciechBarczynski: I think it is a backend implementation driven API design, that makes this API a little bit easier to implement with sacrificing API compactness and making it more boilerplate. IMO doesn't seem like a good idea. Looking at the attached examples, I think that defining `video_n: %InputVideo{input_pad: pad_ref}` each time is a boilerplate solution.

```elixir
# "easy" example without InputVideo struct
  %Scene{
    objects: [
      rounded_1: %Texture{
        input: Map.get(@input_pads_refs, 1),
        transformations: [@corners_rounding]
      },
      rounded_2: %Texture{
        input: Map.get(@input_pads_refs, 2),
        transformations: [@corners_rounding]
      },
      rounded_3: %Texture{
        input: Map.get(@input_pads_refs, 3),
        transformations: [@corners_rounding]
      },
      rounded_4: %Texture{
        input: Map.get(@input_pads_refs, 4),
        transformations: [@corners_rounding]
      },
      final_object: %Overlay{
        @strip_overlay
        | inputs: %{
            top_left: :rounded_1,
            center_left: :rounded_2,
            bottom_left: :rounded_3,
            background: :rounded_4
          },
          resolution: %Resolution{width: 1920, height: 1080}
      }
    ],
    output: :final_object
  }
```

```elixir
# "easy" example with InputVideo struct
  %Scene{
    objects: [
      video_1: %InputVideo{
        input_pad: Map.get(@input_pads_refs, 1)
      },
      video_2: %InputVideo{
        input_pad: Map.get(@input_pads_refs, 2)
      },
      video_3: %InputVideo{
        input_pad: Map.get(@input_pads_refs, 3)
      },
      video_4: %InputVideo{
        input_pad: Map.get(@input_pads_refs, 4)
      },
      rounded_1: %Texture{
        input: :video_1,
        transformations: [@corners_rounding]
      },
      rounded_2: %Texture{
        input: :video_2,
        transformations: [@corners_rounding]
      },
      rounded_3: %Texture{
        input: :video_3,
        transformations: [@corners_rounding]
      },
      rounded_4: %Texture{
        input: :video_4,
        transformations: [@corners_rounding]
      },
      final_object: %Overlay{
        @strip_overlay
        | inputs: %{
            top_left: :rounded_1,
            center_left: :rounded_2,
            bottom_left: :rounded_3,
            background: :rounded_4
          },
          resolution: %Resolution{width: 1920, height: 1080}
      }
    ],
    output: :final_object
  }
```
IMO detecting `Membrane.Pad.ref()` in `input` fields shouldn't be that hard, to justify reducing compactness in API.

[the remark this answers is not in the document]
WojciechBarczynski: > I believe we need to define some kind of a `Video` struct, which would 'connect' a pad reference to a scene object.

Well, not really :)
```elixir
defmodule Membrane.VideoCompositor.Scene do
  @moduledoc """
  Structure representing a top level specification of what is Video Compositor
  supposed to render.
  """

  alias Membrane.VideoCompositor.Scene.Object
  alias Membrane.VideoCompositor.Scene.Object.{InputVideo, Texture}

  @enforce_keys [:objects, :output]
  defstruct @enforce_keys

  @type t :: %__MODULE__{
          objects: [{Object.name(), Object.t()}],
          output: Object.name()
        }

  @spec transform_objects([{Object.name(), Object.t()}]) :: [{Object.name(), Object.t()}]
  def transform_objects(objects) do
    object_names = Enum.map(objects, fn {object_name, _object} -> object_name end)

    pads_refs_to_names =
      Keyword.values(objects)
      |> Enum.map(fn object ->
        case object do
          %Texture{input: input} ->
            input

          layout ->
            Map.get(layout, :inputs)
            |> Map.values()
        end
      end)
      |> List.flatten()
      |> Enum.filter(fn object_name -> not Enum.member?(object_names, object_name) end)
      |> Enum.uniq()
      |> Enum.with_index(fn pad_ref, index -> {pad_ref, String.to_atom("video_#{index}")} end)
      |> Enum.into(%{})

    updated_objects =
      objects
      |> Enum.map(fn {object_name, object} ->
        updated_object =
          case object do
            %Texture{input: input} = texture ->
              %Texture{texture | input: Map.get(pads_refs_to_names, input, input)}

            %{inputs: inputs} = layout ->
              updated_inputs =
                Map.to_list(inputs)
                |> Enum.map(fn {input_name, object_name} ->
                  {input_name, Map.get(pads_refs_to_names, object_name, object_name)}
                end)
                |> Map.new()

              %{layout | inputs: updated_inputs}
          end

        {object_name, updated_object}
      end)
      |> Keyword.new()

    pads_refs_to_names
    |> Map.to_list()
    |> Enum.map(fn {pad_ref, video_name} -> {video_name, %InputVideo{input_pad: pad_ref}} end)
    |> Keyword.new()
    |> Keyword.merge(updated_objects)
  end
end
```

input:
```elixir
input_pads_refs = %{
  1 => :placeholder_for_some_ref_0,
  2 => :placeholder_for_some_ref_1,
  3 => :placeholder_for_some_ref_2,
  4 => :placeholder_for_some_ref_3,
  5 => :placeholder_for_some_ref_4,
  6 => :placeholder_for_some_ref_5
}

one_third = 1.0 / 3.0
sixteen_ninths = 16.0 / 9.0
corners_rounding = %CornersRounding{border_radius: 100}

strip_overlay = %Overlay{
  overlay_spec: %{
    background: %Position{
      top_left_corner: {0.0, 0.0},
      width: 1.0,
      height: 1.0,
      z_value: 0.0
    },
    top_left: %Position{
      top_left_corner: {0.0, 0.0},
      width: sixteen_ninths * one_third,
      height: one_third,
      z_value: 1.0
    },
    center_left: %Position{
      top_left_corner: {0.0, one_third},
      width: sixteen_ninths * one_third,
      height: one_third,
      z_value: 1.0
    },
    bottom_left: %Position{
      top_left_corner: {0.0, 2 * one_third},
      width: sixteen_ninths * one_third,
      height: one_third,
      z_value: 1.0
    }
  },
  inputs: %{},
  resolution: nil
}

objects = [
  rounded_1: %Texture{
    input: Map.get(input_pads_refs, 1),
    transformations: [corners_rounding]
  },
  rounded_2: %Texture{
    input: Map.get(input_pads_refs, 2),
    transformations: [corners_rounding]
  },
  rounded_3: %Texture{
    input: Map.get(input_pads_refs, 3),
    transformations: [corners_rounding]
  },
  rounded_4: %Texture{
    input: Map.get(input_pads_refs, 4),
    transformations: [corners_rounding]
  },
  final_object: %Overlay{
    strip_overlay
    | inputs: %{
        top_left: Map.get(input_pads_refs, 5),
        center_left: :rounded_2,
        bottom_left: Map.get(input_pads_refs, 6),
        background: :rounded_4
      },
      resolution: %Resolution{width: 1920, height: 1080}
  }
]

Scene.transform_objects(objects)
```

output: 
```elixir
[
  video_0: %Membrane.VideoCompositor.Scene.Object.InputVideo{
    input_pad: :placeholder_for_some_ref_0,
    resolution: :input_resolution
  },
  video_1: %Membrane.VideoCompositor.Scene.Object.InputVideo{
    input_pad: :placeholder_for_some_ref_1,
    resolution: :input_resolution
  },
  video_2: %Membrane.VideoCompositor.Scene.Object.InputVideo{
    input_pad: :placeholder_for_some_ref_2,
    resolution: :input_resolution
  },
  video_3: %Membrane.VideoCompositor.Scene.Object.InputVideo{
    input_pad: :placeholder_for_some_ref_3,
    resolution: :input_resolution
  },
  video_5: %Membrane.VideoCompositor.Scene.Object.InputVideo{
    input_pad: :placeholder_for_some_ref_4,
    resolution: :input_resolution
  },
  video_4: %Membrane.VideoCompositor.Scene.Object.InputVideo{
    input_pad: :placeholder_for_some_ref_5,
    resolution: :input_resolution
  },
  rounded_1: %Membrane.VideoCompositor.Scene.Object.Texture{
    input: :video_0,
    transformations: [
      %Membrane.VideoCompositor.Examples.Mock.Transformations.CornersRounding{
        border_radius: 100
      }
    ],
    resolution: :transformed_input_resolution
  },
  rounded_2: %Membrane.VideoCompositor.Scene.Object.Texture{
    input: :video_1,
    transformations: [
      %Membrane.VideoCompositor.Examples.Mock.Transformations.CornersRounding{
        border_radius: 100
      }
    ],
    resolution: :transformed_input_resolution
  },
  rounded_3: %Membrane.VideoCompositor.Scene.Object.Texture{
    input: :video_2,
    transformations: [
      %Membrane.VideoCompositor.Examples.Mock.Transformations.CornersRounding{
        border_radius: 100
      }
    ],
    resolution: :transformed_input_resolution
  },
  rounded_4: %Membrane.VideoCompositor.Scene.Object.Texture{
    input: :video_3,
    transformations: [
      %Membrane.VideoCompositor.Examples.Mock.Transformations.CornersRounding{
        border_radius: 100
      }
    ],
    resolution: :transformed_input_resolution
  },
  final_object: %Membrane.VideoCompositor.Examples.Mock.Layouts.Overlay{
    overlay_spec: %{
      background: %Membrane.VideoCompositor.Examples.Mock.Layouts.Position{
        top_left_corner: {0.0, 0.0},
        width: 1.0,
        height: 1.0,
        z_value: 0.0
      },
      bottom_left: %Membrane.VideoCompositor.Examples.Mock.Layouts.Position{
        top_left_corner: {0.0, 0.6666666666666666},
        width: 0.5925925925925926,
        height: 0.3333333333333333,
        z_value: 1.0
      },
      center_left: %Membrane.VideoCompositor.Examples.Mock.Layouts.Position{
        top_left_corner: {0.0, 0.3333333333333333},
        width: 0.5925925925925926,
        height: 0.3333333333333333,
        z_value: 1.0
      },
      top_left: %Membrane.VideoCompositor.Examples.Mock.Layouts.Position{
        top_left_corner: {0.0, 0.0},
        width: 0.5925925925925926,
        height: 0.3333333333333333,
        z_value: 1.0
      }
    },
    inputs: %{
      background: :rounded_4,
      bottom_left: :video_4,
      center_left: :rounded_2,
      top_left: :video_5
    },
    resolution: %Membrane.VideoCompositor.Scene.Resolution{
      width: 1920,
      height: 1080
    }
  }
]
```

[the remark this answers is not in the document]
WojciechBarczynski: At least the API user doesn't need to specify it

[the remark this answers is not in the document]
WojciechBarczynski: Well, we can also use something like this:
```elixir
defmodule Membrane.VideoCompositor.Scene.Object.InputVideo do
  @moduledoc """
  Structure representing InputVideos wrapper for using Membrane Pad (Element)
  as an object input.
  """
  alias Membrane.Pad

  @enforce_keys [:input_pad]
  defstruct @enforce_keys

  @typedoc """
  Defines videos object, that takes frames from input pad and pass
  them into another objects.
  """
  @type t :: %__MODULE__{
          input_pad: Pad.ref_t()
        }

  @spec from_pad_ref(Pad.ref_t()) :: t()
  def from_pad_ref(ref) do
    %__MODULE__{
      input_pad: ref
    }
  end
end
```

with that, the easy example would look something like this:
```elixir
# with InputVideo struct from pad ref
  %Scene{
    objects: [
      rounded_1: %Texture{
        input: InputVideo.from_pad_ref(Map.get(@input_pads_refs, 1)),
        transformations: [@corners_rounding]
      },
      rounded_2: %Texture{
        input: InputVideo.from_pad_ref(Map.get(@input_pads_refs, 2)),
        transformations: [@corners_rounding]
      },
      rounded_3: %Texture{
        input: InputVideo.from_pad_ref(Map.get(@input_pads_refs, 3)),
        transformations: [@corners_rounding]
      },
      rounded_4: %Texture{
        input: InputVideo.from_pad_ref(Map.get(@input_pads_refs, 4)),
        transformations: [@corners_rounding]
      },
      final_object: %Overlay{
        @strip_overlay
        | inputs: %{
            top_left: :rounded_1,
            center_left: :rounded_2,
            bottom_left: :rounded_3,
            background: :rounded_4
          },
          resolution: %Resolution{width: 1920, height: 1080}
      }
    ],
    output: :final_object
  }
```

pipeline/lib/membrane/video_compositor/pipeline/utility/options.ex
WojciechBarczynski: Ok

lib/membrane/video_compositor/examples/easy.ex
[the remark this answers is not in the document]
WojciechBarczynski: @jerzywilczek IMO `InputVideo.from_pad_ref()` is quite easy to implement/translate to Rust and keeps API concise. What do you think?

lib/membrane/video_compositor/scene.ex
```diff
@@ -2,6 +2,13 @@ defmodule Membrane.VideoCompositor.Scene do
   @moduledoc """
   Structure representing a top level specification of what is Video Compositor
   supposed to render.
+
+  The main part of the Scene are `Membrane.VideoCompositor.Scene.Object`s
+  and interactions between them. There are two kinds of Objects:
+  - `Membrane.VideoCompositor.Scene.Object.Texture` - single input object,
+  taking frames and applying a series of transformations onto it.
+  - `Membrane.VideoCompositor.Scene.Object.Layout` - combining
+  frames from multiple inputs into a single output.
```
WojciechBarczynski, line 11: From this description, it's hard to identify that `Texture` is a "wrapper" struct and `Layout` is a specification for custom object implementation.
```elixir
objects = [
  texture_object_name: %Texture{
    input: :some_input_name,
    transformations: [@cropping, @corners_rounding]
  },
  layout_object_name: %SomeLayoutImplementation{
    @some_layout_impl_spec
    | inputs: %{
        input_name_1: :some_input_name,
        ...
        input_name_n: :some_input_name,
      },
      resolution: %Resolution{width: 1920, height: 1080}
  }
]
```

DominikWolek: I don't think it is necessary here, as moduledocs of Texture and Layouts are one click away. Plus, in that case it would be nice to explain differences between layouts and transformations.

WojciechBarczynski: Ok

WojciechBarczynski (approved):
Beside [this conversation](https://github.com/membraneframework/membrane_video_compositor_plugin/pull/83#discussion_r1133747336) I don't have any other comments. I like this API :)

🚀

lib/membrane/video_compositor/examples/easy.ex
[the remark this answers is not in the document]
jerzywilczek: gowno prawda

varsill (approved):

lib/membrane/video_compositor/scene/transformation.ex
```diff
@@ -0,0 +1,20 @@
+defmodule Membrane.VideoCompositor.Scene.Transformation do
+  @moduledoc """
+  Module representing single input, single output transformations of objects.
+
+  Transformations are single frame input - single frame output objects.
+  Transformations can change frame resolution. Cropping, CornersRounding, ColorFiler,
+  RollToBall, etc. can be implemented as transformations.
+  """
```
WojciechBarczynski, line 8: Using `object` here it might be a little bit confusing since `Transformations` are not `Scene` objects.
```suggestion
defmodule Membrane.VideoCompositor.Scene.Transformation do
  @moduledoc """
  Module representing single-input, single-output frames processing graph nodes
  that apply a single effect on the input frame (e.g. crop them or round corners).

  Transformations are single frame input - single frame output nodes of the processing 
  graph.
  Transformations can change frame resolution. Cropping, CornersRounding, ColorFiler,
  RollToBall, etc. can be implemented as transformations.
  """
```

lib/membrane/video_compositor/scene/transformation.ex
```diff
@@ -0,0 +1,20 @@
...
+
+  @typedoc """
+  Specify how Textures can be defined:
+    - By struct - when transformation can be parametrized with different values
+    e.g. corners round (parametrized with border-radius), cropping, etc.
+    - By module - when there are no reasonable / common use cases of parametrization,
+    and identifying transformation without it is enough e.g. RollToBall - transformation
+    turning input object / frame / video into the ball
+  """
+  @type definition :: struct() | module()
```
WojciechBarczynski, line 18:
```suggestion
  @typedoc """
  Specify how Transformations can be defined:
    - By struct - when transformation can be parametrized with different values
    e.g. corners round (parametrized with border-radius), cropping, etc.
    - By module - when there are no reasonable/common use cases of parametrization,
    and identifying transformation without it is enough e.g. RollToBall - transformation
    turning input object / frame / video into the ball
  """
  @type definition :: struct() | module()
```

lib/membrane/video_compositor/scene/object/texture.ex
```diff
@@ -0,0 +1,35 @@
+defmodule Membrane.VideoCompositor.Scene.Object.Texture do
+  @moduledoc """
+  Texture takes a frame received from Video Compositor objects,
+  applies all transformations and can be passed as an input
+  to other objects.
+  """
```
WojciechBarczynski, line 6:
```suggestion
defmodule Membrane.VideoCompositor.Scene.Object.Texture do
  @moduledoc """
  Texture takes a frame received from Video Compositor objects,
  applies all transformations and can be passed as an input
  to other objects.
  
  Basically wraps multiple single-input, single-output processing graph nodes.
  """
```

lib/membrane/video_compositor/scene/object.ex
```diff
@@ -0,0 +1,40 @@
...
+  alias Membrane.VideoCompositor.Scene.Object.{InputVideo, Layout, Texture}
+
+  @typedoc """
+  Objects are renderable entities in VC, that can serve as input for other
+  objects or as an output of the video.
+
+  They are either Texture structs or structs defining Layouts
+  following Layout.t() definition.
+  """
+  @type t :: Layout.t() | Texture.t() | InputVideo.t()
```
WojciechBarczynski, line 15:
```suggestion
  They are either Texture structs, structs defining Layouts
  following Layout.t() definition or InputVideo structs.
  """
  @type t :: Layout.t() | Texture.t() | InputVideo.t()
```

lib/membrane/video_compositor/scene/object/layout.ex
```diff
@@ -0,0 +1,42 @@
...
+
+  Layouts can take multiple renderable objects as input and combine
+  them into one frame. Fadings, Grids, Overlays, Transitions, etc.
+  can be defined as Layouts.
+
+  Basically it's a multi-input, single-output node in processing graph.
+  """
+  alias Membrane.VideoCompositor.Scene.{Object, Resolution}
+
+  # @enforced_keys [:inputs, :resolution]
```
WojciechBarczynski, line 13: Is it necessary here? `@type t` definition already specifies that

lib/membrane/video_compositor/scene.ex
```diff
@@ -0,0 +1,25 @@
...
+
+  The main part of the Scene are `Membrane.VideoCompositor.Scene.Object`s
+  and interactions between them. There are three kinds of Objects:
+  - `Membrane.VideoCompositor.Scene.Object.Texture` - single input object,
+  taking frames and applying a series of transformations onto it.
+  - `Membrane.VideoCompositor.Scene.Object.Layout` - combining
+  frames from multiple inputs into a single output.
+  - `Membrane.VideoCompositor.Scene.Object.InputVideo` - which maps
+  an input pad of element into Scene object.
+  """
```
WojciechBarczynski, line 14: NIT - this order better follows the usual definition of objects in the scene.
```suggestion
  The main part of the Scene are `Membrane.VideoCompositor.Scene.Object`s
  and interactions between them. There are three kinds of Objects:
   - `Membrane.VideoCompositor.Scene.Object.InputVideo` - which maps
  an input pad of element into Scene object.
  - `Membrane.VideoCompositor.Scene.Object.Texture` - single input object,
  taking frames and applying a series of transformations onto it.
  - `Membrane.VideoCompositor.Scene.Object.Layout` - combining
  frames from multiple inputs into a single output.
  """
```

WojciechBarczynski: Moreover, I think the second part should rather be in `@type t` `@typedoc`

lib/membrane/video_compositor/scene.ex
```diff
@@ -0,0 +1,25 @@
...
+
+  alias Membrane.VideoCompositor.Scene.Object
+
+  @enforce_keys [:objects, :output]
+  defstruct @enforce_keys
+
+  @type t :: %__MODULE__{
+          objects: [{Object.name(), Object.t()}],
+          output: Object.name()
+        }
```
WojciechBarczynski, line 24: Maybe we can leave a note here, that if all `object names` are atoms, `objects` can simply be a keyword list

WojciechBarczynski (approved):
Overall I like this API. I left a few suggestions, but nothing major.

WojciechBarczynski:
> _ No description provided. _

Maybe living here sth like this would make it easier to understand for VC newcomers:

```
Here is some key API info:

- All VC processing is defined by the Scene struct
- There are three types of Objects (renderable things): InputVideos, Textures and Layouts
- InputVideos wrap input pads as a processing graph node
- Texture objects are single input and single output nodes in a processing graph
- Layout objects are multiple inputs and a single output nodes in a processing graph
- Objects take other Objects as inputs
- All Objects can be imagined as things taking the simple image(s), processing them, and returning new image
- User of VC can create custom Transformations (applied to textures) and Layouts (hopefully writing only a little of Elixir, Rust, and shader code
```

lib/membrane/video_compositor/scene.ex
WojciechBarczynski: @Janix4000 What do you think?

jerzywilczek: I don't understand what you're suggesting @WojciechBarczynski

lib/membrane/video_compositor/scene/transformation.ex
```diff
@@ -0,0 +1,23 @@
+defmodule Membrane.VideoCompositor.Scene.Transformation do
+  @moduledoc """
+  Module representing single-input, single-output frames processing graph nodes
+  that apply a single effect on the input frame (e.g. crop them or round corners).
+
+  Transformations are single frame input - single frame output nodes of the processing
+  graph.
```
WojciechBarczynski, line 7:
```suggestion
defmodule Membrane.VideoCompositor.Scene.Transformation do
  @moduledoc """
  Module representing single-input, single-output frames processing graph nodes
  that apply a single effect on the input frame (e.g. crop them or round corners).
```
pleonasm

lib/membrane/video_compositor/scene/object.ex
```diff
@@ -0,0 +1,40 @@
...
+  Objects can be assigned to names and identified
+  at other objects as inputs based on assigned names
+  """
+  @type name :: tuple() | atom()
+
+  @typedoc """
+  Defines how the input of an object can be specified
+  in Video Compositor.
+  """
+  @type input :: name()
```
jerzywilczek, line 29: I think the `input` type can be removed and replaced with `name`

lib/membrane/video_compositor/scene.ex
WojciechBarczynski: My original suggestion was just to leave in `@typedoc` info, that if all names of objects are simply atoms, then we can define `objects: some_keyword_list` as an field in the `scene` struct (where `Keyword.keyword?(some_keyword_list)` is `true`). Then Dominik suggested removing `tuple` from the `Object.name` type definition, such that we can always use a keyword list here (at least from what have I understood)

Janix4000: There is no significant advantage to restricting users to use only atoms as object ids. From our perspective, there is no difference in the types of ids. From the users' perspective, it is very convenient to use tuples of atom names and integer indices instead of generated atoms. Additionally, it is not recommended to generate atoms in runtime.

jerzywilczek: it actually makes no difference in the code whatsoever, we will hold names as `rustler::Term`s on the rust side, which means anything can be a name

WojciechBarczynski: Ok

lib/membrane/video_compositor/scene/object/input_video.ex
```diff
@@ -0,0 +1,17 @@
...
+
+  @enforce_keys [:input_pad]
+  defstruct @enforce_keys
+
+  @typedoc """
+  Defines pad wrapper.
+  """
+  @type t :: %__MODULE__{
+          input_pad: Pad.ref_t()
+        }
```
jerzywilczek, line 16: I don't know if we shouldn't hold the stream format here

jerzywilczek (approved):

lib/membrane/video_compositor/scene/object/input_video.ex
WojciechBarczynski: We shouldn't - input stream format can change with membrane callback, meaning we would need to enforce updating the scene with it, even when the VC output image doesn't change

WojciechBarczynski: Because of that, binding the `InputVideo` output resolution to a certain value also doesn't seem like a good idea for me either.

lib/membrane/video_compositor/examples/easy.ex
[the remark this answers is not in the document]
WojciechBarczynski: Po prostu nie umiesz

lib/membrane/video_compositor/scene/object.ex
```diff
@@ -0,0 +1,34 @@
...
+  """
+  @type t :: Layout.t() | Texture.t() | InputVideo.t()
+
+  @typedoc """
+  Defines how an object can be referenced in Scene.
+
+  Objects can be assigned to names and identified
+  at other objects as inputs based on assigned names
+  """
+  @type name :: tuple() | atom()
```
WojciechBarczynski, line 23: Since it will be kept in `Term` in Rust, which implements `hash` and `equals`, AFAIK we can even use `any()` type here (@jerzywilczek correct me if I'm wrong). 
```suggestion
  @type name :: any()
```

DominikWolek: I would still leave this type restricted. Especially that it's identical in `Membrane.Core` https://hexdocs.pm/membrane_core/Membrane.Element.html#t:name_t/0

lib/membrane/video_compositor/scene/object/input_video.ex
jerzywilczek: That's true, but this makes me a bit unsure about the way of handling stream format changes at the boundary between the queue and the compositor

lib/membrane/video_compositor/examples/mock/layouts/position.ex
```diff
@@ -0,0 +1,20 @@
...
+  @typedoc """
+  Float in [0, 1] range. Widely used in graphic rendering, to determine distance, color etc.
+  """
+  @type unit_range :: float()
+
+  @type t :: %__MODULE__{
+          top_left_corner: {unit_range(), unit_range()},
+          width: unit_range() | :auto,
+          height: unit_range() | :auto,
+          z_value: unit_range()
```
bblaszkow06, line 18: Why are these floats? It seems quite unusual to define either of these as floats. Additionally, what do the `width` and `height` refer to? The scale of the input or percentage of available width or height? It seems ambiguous

WojciechBarczynski: Since input texture resolution (input from some membrane pad) might change, depending on e.g. network conditions, defining cropping parameters in relative coordinates (like [0, 1] range used here) can deal with such issues.

bblaszkow06 (approved):

PR #84, opened 2023-01-30
Fix removing pads from compositor bin

Karolk99:
After this fix you'll be able to unlink pads from bin. Right now it's not possible because `framerate converter` is never removed. 
Also I recommend to include pad removing and caps that change over time in tests.


jerzywilczek (approved):
:rocket:

PR #86, opened 2023-02-08
Add a better picture comparison function for tests.

jerzywilczek:
We now use mean square error to compare pictures in tests which compare a reference picture with a generated one, instead of a binary 'are these two pixels the same' check.

### DIFF
--- a/test/support/utility.ex
+++ b/test/support/utility.ex
@@ -258,17 +258,20 @@ defmodule Membrane.VideoCompositor.Test.Support.Utils do
   Compare contents of two raw files, succeeding when `different_bytes / length < allowed_error`
   """
   @spec compare_contents_with_error(binary(), binary(), float()) :: boolean()
-  def compare_contents_with_error(output_path, reference_path, allowed_error \\ 0.003) do
+  def compare_contents_with_error(output_path, reference_path, allowed_error \\ 1.5) do
     {:ok, reference_file} = File.read(reference_path)
     {:ok, output_file} = File.read(output_path)
 
     assert byte_size(reference_file) == byte_size(output_file)
 
-    errors =
+    square_errors =
       Enum.zip(:binary.bin_to_list(reference_file), :binary.bin_to_list(output_file))
-      |> Enum.count(fn {b1, b2} -> b1 != b2 end)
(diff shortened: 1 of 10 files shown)


mat-hek (approved):

mat-hek:
tests pass on my mac ✅

test/support/utility.ex
```diff
@@ -258,17 +258,20 @@ defmodule Membrane.VideoCompositor.Test.Support.Utils do
   Compare contents of two raw files, succeeding when `different_bytes / length < allowed_error`
   """
   @spec compare_contents_with_error(binary(), binary(), float()) :: boolean()
-  def compare_contents_with_error(output_path, reference_path, allowed_error \\ 0.003) do
+  def compare_contents_with_error(output_path, reference_path, allowed_error \\ 1.5) do
```
WojciechBarczynski, line 261: Nit: `allowed_mse` instead of `allowed_error` would be a little bit more descriptive.

WojciechBarczynski (approved):
One small Nit. You can change it if you want to.

test/support/utility.ex
```diff
@@ -255,20 +255,23 @@ defmodule Membrane.VideoCompositor.Test.Support.Utils do
   end
 
   @doc """
-  Compare contents of two raw files, succeeding when `different_bytes / length < allowed_error`
+  Compare contents of two raw files, succeeding when `mse < allowed_error`
   """
   @spec compare_contents_with_error(binary(), binary(), float()) :: boolean()
-  def compare_contents_with_error(output_path, reference_path, allowed_error \\ 0.003) do
+  def compare_contents_with_error(output_path, reference_path, allowed_error \\ 1.5) do
```
WojciechBarczynski, line 261: I just recalled, that since we are changing how the ref video is compared with the test output video, we should also remove platform-based separation in [`test/texture_transformations_test.exs`](https://github.com/membraneframework/membrane_video_compositor_plugin/blob/master/test/texture_transformations_test.exs#:~:text=to%20different%20references.-,%40tag%20linux%3A%20true,-test%20%221%20frame). Linux/mac separation shouldn't be required anymore.

WojciechBarczynski (changes requested):

PR #90, opened 2023-02-23
Add an initial design of the rust part of the new compositor API.

jerzywilczek:


WojciechBarczynski (approved):

jerzywilczek:
@WojciechBarczynski @Janix4000 can you give an opinion about the open question at the end of the file and in the `MockScene` struct?

native/membrane_videocompositor/src/new_compositor.rs
```diff
@@ -0,0 +1,62 @@
...
+        _pts: Timestamp,
+    ) -> Result<Frame, ComposeError> {
+        todo!()
+    }
+}
+
+// Open questions:
+//  - do we need pts for each frame? I think we don't and components that use time information
+//    should use the timestamp of the whole frame as their time reference, but I would love to
+//    hear some opinions on that
```
WojciechBarczynski, line 62: I currently can't come up with a component in which having individual frames pts would be beneficial. Moreover, I think handling timestamp normalization in the queue would simplify creating new components and it's more natural for me to handle that there.

DominikWolek: I reckon, that in one of the earliest brainstorming of VC we decided components should not rely on the pts. Instead, if they ever need one, VC can always provide them percentages or something similar.

DominikWolek: So I second @WojciechBarczynski opinion

native/membrane_videocompositor/src/new_compositor.rs
```diff
@@ -0,0 +1,62 @@
+use std::sync::{Arc, Mutex};
```
DominikWolek, line 1: `new_compositor.rs` is a little iffy for be, but I understand it's going to change soon

DominikWolek (approved):
image

native/membrane_videocompositor/src/new_compositor.rs
jerzywilczek: > I reckon, that in one of the earliest brainstorming of VC we decided components should not rely on the pts. Instead, if they ever need one, VC can always provide them percentages or something similar.

This seems complicated, I'm not sure how we would pass this time information, since I think calculating it would be the responsibility of the queue. Any ideas?

native/membrane_videocompositor/src/new_compositor.rs
jerzywilczek: yes, but I don't know what to rename it to.

jerzywilczek:
Closing due to this being more of a brainstorm PR, not intended to actually merge.

PR #94, opened 2023-02-27
Fix CircleCI config

WojciechBarczynski:


.circleci/config.yml
```diff
@@ -56,16 +56,29 @@ jobs:
...
+              only: /v.*/
+      - elixir_test_with_software_rendering:
+          filters:
+            <<: *filters
+      - elixir/lint:
+          filters:
+            <<: *filters
+      - rust_lint:
+          filters:
+                <<: *filters
```
varsill, line 71: there are more tabs there than in the cases above - I would level it :p

varsill (approved):

PR #95, opened 2023-02-28
Simplified scene API design

WojciechBarczynski:
Original Scene API seems not to follow the Membrane motto `Make simple things easy and hard things possible`, at least in my opinion. This is my take on simplifying it while trying to tackle the problem of low-level canvases implementations (which raised some questions in last week).


WojciechBarczynski:
Ported to #96

PR #98, opened 2023-04-05
Ms 422 design scene graph on the rust side

jerzywilczek:
This is getting to be more and more final.

Things still left to do (that I can recall should be done):
 - [x] Document all structs, types and functions
 - [x] Add e2e tests in the form of parsing the scene design examples
 - [x] Refactor the code in some places, mostly the `encode` functions in Elixir

I'm especially open to proposals for changing the names of various structs.


lib/membrane/video_compositor/scene/object/layout.ex
```diff
@@ -10,11 +10,62 @@ defmodule Membrane.VideoCompositor.Scene.Object.Layout do
...
+  end
+
   @typedoc """
-  Type of a map defining on how to map internal layout's ids
-  to Scene objects
+  A layout-internal identifier, which the layout can use to determine
+  the way of rendering specific input objects. See `t:inputs/0` for more info.
   """
-  @type inputs :: %{any() => Object.input()}
+  @type internal_name :: Object.name()
```
WojciechBarczynski, line 51: In some cases `atom` in `{atom(), number}` tuple feels unnecessary. e.g. repeating `:input` in your tests
```suggestion
  @type internal_name :: Object.name() | integer() | float()
```

test/support/mock/layouts/grid.ex
```diff
@@ -19,4 +20,9 @@ defmodule Membrane.VideoCompositor.Examples.Mock.Layouts.Grid do
           },
           resolution: Resolution.t()
         }
+
+  @impl true
+  def encode(_grid) do
+    0
+  end
```
WojciechBarczynski, line 27: It's weird. Same for other mocks.

test/support/mock/transformations/rotate.ex
```diff
@@ -0,0 +1,17 @@
...
+  defstruct @enforce_keys
+
+  @type t :: %__MODULE__{
+          degrees: non_neg_integer()
+        }
+
+  @impl true
+  def encode(_rotate) do
+    2137
+  end
```
WojciechBarczynski, line 16: ...
```suggestion
  @impl true
  def encode(_rotate) do
    69
  end
```

native/membrane_videocompositor/src/elixir_bridge/elixir_structs.rs
```diff
@@ -136,3 +140,710 @@ impl From<ElixirCropping> for Box<dyn TextureTransformation> {
...
+    fn mentioned_names(&self) -> Vec<&Name> {
+        let mut names = vec![&self.input];
+
+        if let TextureOutputResolution::Name(name) = &self.resolution {
+            names.push(name);
+        }
+
+        names
+    }
+}
```
WojciechBarczynski, line 171: Since we implement the same `mentioned_names` function in `Texture` and `Layout`, I think that we should define `Trait` for that and implement it here

native/membrane_videocompositor/src/elixir_bridge/elixir_structs.rs
```diff
@@ -136,3 +140,710 @@ impl From<ElixirCropping> for Box<dyn TextureTransformation> {
...
+
+impl Object {
+    fn mentioned_names(&self) -> Vec<&Name> {
+        match self {
+            Object::Layout(layout) => layout.mentioned_names(),
+            Object::Texture(texture) => texture.mentioned_names(),
+            Object::Video(_) => Vec::new(),
+        }
+    }
+}
```
WojciechBarczynski, line 242: Use trait for that

native/membrane_videocompositor/src/elixir_bridge/elixir_structs.rs
```diff
@@ -136,3 +140,710 @@ impl From<ElixirCropping> for Box<dyn TextureTransformation> {
...
+            panic!("unexpected scene structure");
+        };
+        assert_eq!(transformation, 1);
+
+        let Node::Transformation { transformation, .. } = **previous else {
+            panic!("unexpected scene structure");
+        };
+        assert_eq!(transformation, 0);
+    }
+}
```
WojciechBarczynski, line 849: This module does literary everything. Maps elixir structs, handle rustler weird stuff, validate scene, defines validation and parsing errors ... Basically, responsibilities separation doesn't exist here at all. It's just a `one module to rule them all`. I would separate it, such that:
- dealing with rustler stuff is handled separately
- validation is separated from translation 

download (2)

native/membrane_videocompositor/src/elixir_bridge/elixir_structs.rs
```diff
@@ -136,3 +140,710 @@ impl From<ElixirCropping> for Box<dyn TextureTransformation> {
...
+        let objects = self
+            .objects
+            .iter()
+            .map(|(name, object)| (name, object))
+            .collect::<HashMap<_, _>>();
+
+        Self::contains_cycle(&objects, &self.output)?;
+
+        Ok(())
+    }
```
WojciechBarczynski, line 385: I don't think that module named `elixir_bridge/elixir_structs` should be responsible for scene validation.

native/membrane_videocompositor/src/elixir_bridge/elixir_structs.rs
```diff
@@ -136,3 +140,710 @@ impl From<ElixirCropping> for Box<dyn TextureTransformation> {
...
+                        .collect::<Result<HashMap<_, _>, _>>()?;
+
+                    let result = Arc::new(Node::Layout {
+                        resolution: resolution.clone(),
+                        implementation: *implementation,
+                        inputs,
+                    });
+                    nodes.insert(name.clone(), result.clone());
+                    Ok(result)
+                }
```
WojciechBarczynski, line 462: With separate functions for that logic, it would be more readable.

WojciechBarczynski (changes requested):

test/support/mock/transformations/rotate.ex
jerzywilczek: absolutely not

lib/membrane/video_compositor/scene/object/layout.ex
```diff
@@ -10,11 +10,62 @@ defmodule Membrane.VideoCompositor.Scene.Object.Layout do
...
+    A rustler-friendly version of the inputs map. Keep in mind that this maps
+    __internal names__ => __scene object names__.
+    """
+    @type inputs :: %{internal_name() => RFObject.name()}
+    @type output_resolution :: {:resolution, Resolution.t()} | {:name, Object.name()}
+
+    # in a more 'final' product this should be some kind of a layout identifier.
+    # I thought of making this a UUID that would correspond to an implementation
+    # on the rust side, but layout names would work fine too.
+    @type rust_representation :: integer()
```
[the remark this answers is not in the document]
jerzywilczek: this is the closest thing to what will actually be sent, I'm going to leave this in.

lib/membrane/video_compositor/scene/transformation.ex
```diff
@@ -19,5 +19,19 @@ defmodule Membrane.VideoCompositor.Scene.Transformation do
...
   @type definition :: struct() | module()
-  @type name :: tuple() | atom()
+
+  @typedoc """
+  A rust representation of the transformation, passed through elixir in an opaque way.
+  """
+  # in a more 'final' product this should be some kind of a transformation identifier.
+  # I thought of making this a UUID that would correspond to an implementation
+  # on the rust side, paired with parameters.
+  @type rust_representation :: integer()
```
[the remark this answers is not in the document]
jerzywilczek: same as above

native/membrane_videocompositor/src/elixir_bridge/elixir_structs.rs
jerzywilczek: it doesn't really make sense, since it wouldn't simplify any code. This would be a good solution if we wanted `Object` to not be an `enum` and hold a Box<dyn Sth> instead. Using a trait doesn't fit here.

native/membrane_videocompositor/src/elixir_bridge/elixir_structs.rs
jerzywilczek: answered above

native/membrane_videocompositor/src/elixir_bridge/elixir_structs.rs
jerzywilczek: I actually think it would be less readable, since the code for recursing into different variants would be in separate places in the file. I find it nicer to read like it is now.

lib/membrane/video_compositor/scene/object/layout.ex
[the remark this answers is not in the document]
WojciechBarczynski: Ok

native/membrane_videocompositor/src/elixir_bridge/elixir_structs.rs
jerzywilczek: i split it

lib/membrane/video_compositor/scene/object/layout.ex
WojciechBarczynski, line 34: Defining `rust_representation` `type` twice in `Layout` and `RFLayout` seems unnecessary. Moreover those types are inconsistent with each other since one is defined as `integer()` and the other on is defined as `non_neg_integer()`

lib/membrane/video_compositor/scene/object/texture.ex
```diff
@@ -34,4 +56,27 @@ defmodule Membrane.VideoCompositor.Scene.Object.Texture do
...
+  @spec encode(t()) :: RustlerFriendly.t()
+  def encode(texture) do
+    encoded_transformations =
+      texture.transformations
+      |> Enum.map(fn transformation ->
+        case transformation do
+          %module{} -> module.encode(transformation)
+          module -> module.encode(transformation)
+        end
+      end)
```
WojciechBarczynski, line 72: I would define this logic in the `Membrane.VideoCompositor.Transformation` module and call it from here since it depends on the `type` defined there

WojciechBarczynski (changes requested):
Review II 
part I - Elixir

lib/membrane/video_compositor/scene/object/layout.ex
```diff
@@ -37,4 +88,43 @@ defmodule Membrane.VideoCompositor.Scene.Object.Layout do
...
+      resolution: encoded_resolution,
+      implementation: rust_representation
+    }
+  end
+
+  @spec encode_internal_name(internal_name()) :: RustlerFriendly.internal_name()
+  defp encode_internal_name(name) do
+    Object.encode_name(name)
+  end
 end
```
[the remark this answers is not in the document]
jerzywilczek: it is not though. our `encode/1` is actually `Layout.encode/1`, and their function would be `UserDefinedStruct.encode/1`. Our function actually calls theirs, in line 109. Given that, I think we should keep it this way.

lib/membrane/video_compositor/scene/object/texture.ex
jerzywilczek: I agree, but that makes the `Transformation` module the same as the `Layout` module, which you have taken issue with in the comment above

native/membrane_videocompositor/src/elixir_bridge/elixir_structs.rs
```diff
@@ -136,3 +140,328 @@ impl From<ElixirCropping> for Box<dyn TextureTransformation> {
...
+    height: u32,
+}
+
+#[derive(Debug, rustler::NifStruct)]
+#[module = "Membrane.VideoCompositor.Scene.Object.Texture.RustlerFriendly"]
+pub struct Texture {
+    pub input: ObjectName,
+    // this is a placeholder only temporarily, until we figure out encoding transformations
+    // then, it will be something like `Arc<dyn Transformation>`
+    pub transformations: Vec<ImplementationPlaceholder>,
```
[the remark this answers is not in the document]
jerzywilczek: It obviously will be changed when we implement user transformations, for now i'd rather be explicit about this being a placeholder here.

native/membrane_videocompositor/src/elixir_bridge/elixir_structs.rs
jerzywilczek: notice how clippy doens't mark it as bad. this level of indentation is rather normal in conventional rust.

WojciechBarczynski: It is a different language, but the 60 lines long match is unreadable. There are a few problems I'm noticing here:
1. This `parse_object` function is defined in `Scene` when it depends on the `Object` structure. This violates the single responsibility principle. When we change the definition of `Object`, we should expect only to change the `Object` implementation, but here we have actually to change the `Scene` struct. This creates unnecessary dependency, complicates the codebase, and makes it harder to extend in the future. I would move it to `Object` `impl`
2. There is an inconsistency between this giant pattern match, which handles the whole logic for parsing all types of objects, and the `mantioned_names` pattern match, which does it with separate functions. This pattern match will even grow bigger once we add new objects / change the current one.
3. Chandling those conversions in a single place, means that we have to modify it whenever we change the single one. By splitting this logic to separate functions for each type of `Object`, we only have to modify the code responsible for the appropriate `Object` implementation.

Therefore, I think that:
1. `parse_object` should be moved to `Object`
2. Each object type should handle its parsing inside their own implementation

native/membrane_videocompositor/src/scene.rs
```diff
@@ -0,0 +1,37 @@
...
+    Video { pad: PadRef },
+}
+
+/// A scene represents the full pipeline from video inputs (that come via Membrane pads)
+/// to outputs. It holds a reference to a [Node] which represents the final frame that will be rendered,
+/// which will then be returned to elixir and sent out to the next Membrane element.
+#[derive(Debug)]
+pub struct Scene {
+    pub final_node: Arc<Node>,
+}
```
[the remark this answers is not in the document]
WojciechBarczynski: I thought about it for a while, and I think that with this you can perceive it as sth similar to a linked list.  I guess it makes some sense.

native/membrane_videocompositor/src/elixir_bridge/elixir_structs.rs
WojciechBarczynski: 3. Each of these match arms repeats this logic:
```rust
                    nodes.insert(name.clone(), current.clone());
                    Ok(node)
```

native/membrane_videocompositor/src/elixir_bridge/scene_validation.rs
```diff
@@ -0,0 +1,392 @@
...
+                if set.contains(input_pad) {
+                    return Err(SceneParsingError::DuplicatePadReferences(input_pad.clone()));
+                }
+
+                set.insert(input_pad);
+            }
+        }
+
+        Ok(())
+    }
```
WojciechBarczynski, line 56: [NIT] The `Set` name is too vague, it doesn't help to understand the code. Use `input_pads` or a similar name for this.

native/membrane_videocompositor/src/elixir_bridge/scene_validation.rs
```diff
@@ -0,0 +1,392 @@
...
+        if *difference[0] != final_object_name {
+            return Err(SceneParsingError::UnusedObject((*difference[0]).clone()));
+        }
+
+        if difference.len() > 1 {
+            return Err(SceneParsingError::UnusedObject((*difference[1]).clone()));
+        }
+
+        Ok(())
+    }
```
WojciechBarczynski, line 112: All names are used somehow. `final_object` is just referenced as `filan_object`. The naming convention here might be confusing.
```suggestion
    fn check_for_unused_objects(
        objects: &[(ObjectName, Object)],
        final_object_name: &ObjectName,
    ) -> Result<(), SceneParsingError> {
        let defined_names = objects.iter().map(|(name, _)| name).collect::<HashSet<_>>();

        let mut used_names = objects
            .iter()
            .flat_map(|(_, object)| object.mentioned_names())
            .collect::<HashSet<_>>();

        if used_names.contains(final_object_name) {
            return Err(SceneParsingError::CycleDetected);
        } else {
            used_names.insert(final_object_name);
        }

        let unused_names = defined_names.difference(&used_names).collect::<Vec<_>>();

        if !unused_names.is_empty() {
            return Err(SceneParsingError::UnusedObject((*unused_names[0]).clone()));
        }

        Ok(())
    }
```

native/membrane_videocompositor/src/elixir_bridge/scene_validation.rs
```diff
@@ -0,0 +1,392 @@
...
+
+            match &objects[name] {
+                Object::Video(_) => Ok(()),
+                Object::Texture(Texture { input, .. }) => visit(input, objects, visited),
+                Object::Layout(Layout { inputs, .. }) => inputs
+                    .values()
+                    .map(|name| visit(name, objects, visited))
+                    .collect::<Result<Vec<()>, SceneParsingError>>()
+                    .map(|_| ()),
+            }?;
```
WojciechBarczynski, line 145: This will silently break if we change how we define neighbor nodes. I think that a separate function like `get_previous_nodes` would make it harder to forget to update that.

native/membrane_videocompositor/src/elixir_bridge/scene_validation.rs
```diff
@@ -0,0 +1,392 @@
...
+        }
+
+        names
+    }
+}
+
+impl Scene {
+    fn check_for_duplicate_pad_refs(
+        objects: &[(ObjectName, Object)],
+    ) -> Result<(), SceneParsingError> {
```
WojciechBarczynski, line 42: Since we check for that, we should specify in the `InputVideo` docs, that pads can only be used once.

native/membrane_videocompositor/src/elixir_bridge/elixir_structs.rs
```diff
@@ -136,3 +140,328 @@ impl From<ElixirCropping> for Box<dyn TextureTransformation> {
...
+    type Error = SceneParsingError;
+
+    fn try_into(self) -> Result<crate::scene::Scene, Self::Error> {
+        self.validate()?;
+
+        let node = self.convert_to_graph()?;
+
+        Ok(crate::scene::Scene { final_node: node })
+    }
+}
```
WojciechBarczynski, line 367: Shouldn't this logic be in `scene_validation.rs`? If it is possible to have multiple modules with `rustler` translation, I would move that there.

WojciechBarczynski (changes requested):
Review II - part 2

native/membrane_videocompositor/src/elixir_bridge/elixir_structs.rs
WojciechBarczynski: Well, it's still weird to have WIP `Scene` modules in the same module with working stuff like `ElixirBaseVideoPlacement`, specific to the current VC version.

jerzywilczek: perhaps the work on the new compositor should be moved away from master into a different branch

native/membrane_videocompositor/src/elixir_bridge/elixir_structs.rs
```diff
@@ -136,3 +140,328 @@ impl From<ElixirCropping> for Box<dyn TextureTransformation> {
...
+                            previous: current,
+                            transformation,
+                            resolution: resolution.clone(),
+                        });
+                    }
+
+                    nodes.insert(name.clone(), current.clone());
+
+                    Ok(current)
+                }
```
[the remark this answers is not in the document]
jerzywilczek: I was under the impression that if a resolution is specified explicitly (e.g. you just pass in 1920x1080) then no matter what the transformation does this should be the output resolution of this transformation. The two other ways of specifying the resolution are either to specify it as a `:transformed_input_resolution` or as a name of another object. In these cases the information copied here is either `Resolution::Name(ObjectName)` or `Resolution::TransformedInputResolution`, both of which indicate that actual numeric resolutions weren't resolved yet.

lib/membrane/video_compositor/scene/object.ex
```diff
@@ -5,6 +5,23 @@ defmodule Membrane.VideoCompositor.Scene.Object do
...
+    @type name ::
+            {:atom, binary()}
+            | {:atom_pair, binary(), binary()}
+            | {:atom_num, binary(), non_neg_integer()}
+
+    @type t :: {:layout, RFLayout.t()} | {:texture, RFTexture.t()} | {:video, RFInputVideo.t()}
+
+    @type object_output_resolution :: RFTexture.output_resolution() | RFLayout.output_resolution()
+  end
+
```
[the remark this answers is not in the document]
jerzywilczek: elixir doesn't allow this, submodules defined in the same file as other code must be defined before they're used in the other code. its such a silly language xd

native/membrane_videocompositor/src/elixir_bridge/elixir_structs.rs
jerzywilczek: How does it make it harder to extend in the future? any change in the definition of any of these structs will force you (by erroring out during compilation) to change the code here as well. This will happen automatically, you will be brought to this code by the compiler. The definitions are not in some remote place at the other end of the project. They're in the same module.

I'd say the SRP in rust applies more to modules than to individual structs. Also, quite often data-oriented design is preferred over object oriented programming patterns in rust. This means that you treat individual structs more as data which is open to being modified and processed and less as self-contained objects that talk to each other.

The inconsistency between this function and `mentioned_names` stems from the fact that the functionality in `mentioned_names` could be used somewhere else in the module (I think this even is the case), and can be derived from the object alone, without any external context. On the other hand, I cannot come up with a way to utilize `parse_object` in a different part of this module and parsing doesn't really make sense to be done without the additional information contained in the scene. I believe (at least from the data-oriented design point of view) since all of the code does a very specific thing, which is almost certainly useless for the rest of the project, it makes sense to put all of it in one spot where it is used.

The third point doesn't make sense for me. Yes, we have to modify this code whenever one of these objects changes. I don't see why it would be nicer to edit the code for parsing `Texture` in `Texture::parse` rather the part of the match responsible for handling `Texture`. There is no difference, you only have to edit the code responsible for `Texture` in both cases.

Last part: I opened the repo for the rustc compiler, which is, to my knowledge, the largest rust project in existence. Hundreds of people work on it together daily. looked through <10 files and found [this](https://github.com/rust-lang/rust/blob/e49122fb1ca87a6c3e3c22abb315fc75cfe8daed/compiler/rustc_ast_passes/src/ast_validation.rs#L806). I know this is a different case than what we have here, I just wanted to demonstrate that very long match statements matching over enums being located outside of these enums' `impl` blocks are considered rather normal in rust. I wouldn't call a 60 lines long match statement 'unreadable', since these are commonplace even in extremely large projects, which tend to place strong emphasis on the code being readable to all contributors.

I will move this match to `Object` however, because I don't think the negative consequences of doing that are significant enough to discuss further.

lib/membrane/video_compositor/scene/object.ex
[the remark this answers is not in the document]
jerzywilczek: ok nvm i did it

lib/membrane/video_compositor/scene.ex
```diff
@@ -4,9 +4,22 @@ defmodule Membrane.VideoCompositor.Scene do
...
+    alias Membrane.VideoCompositor.Scene.Object.RustlerFriendly, as: RFObject
+
+    @type t :: %__MODULE__{
+            objects: [{RFObject.name(), RFObject.t()}],
+            output: RFObject.name()
+          }
+
+    @enforce_keys [:objects, :output]
+    defstruct @enforce_keys
+  end
```
WojciechBarczynski, line 21: [NIT] Move it below API specification (under `@type t` definition)

native/membrane_videocompositor/src/elixir_bridge/elixir_structs.rs
```diff
@@ -136,3 +142,260 @@ impl From<ElixirCropping> for Box<dyn TextureTransformation> {
...
+}
+
+/// A reference to a Membrane.Pad
+pub type PadRef = String;
+
+#[derive(Debug, rustler::NifStruct)]
+#[module = "Membrane.VideoCompositor.Scene.Object.InputVideo.RustlerFriendly"]
+pub struct InputVideo {
+    pub input_pad: PadRef,
+}
```
WojciechBarczynski, line 310: [NIT] move it closer to other objects definitions

WojciechBarczynski:
Update PR description

lib/membrane/video_compositor/scene/object/texture.ex
```diff
@@ -34,4 +35,43 @@ defmodule Membrane.VideoCompositor.Scene.Object.Texture do
...
+  end
+
+  @doc false
+  # Encode the texture to a Texture.RustlerFriendly in order to prepare it for
+  # the rust conversion.
+  @spec encode(t()) :: RustlerFriendly.t()
+  def encode(texture) do
+    encoded_transformations =
+      texture.transformations
+      |> Enum.map(&Transformation.encode(&1))
```
Janix4000, line 67:
```suggestion
      |> Enum.map(&Transformation.encode/1)
```

lib/membrane/video_compositor/scene/object/texture.ex
```diff
@@ -34,4 +35,43 @@ defmodule Membrane.VideoCompositor.Scene.Object.Texture do
...
+
+    encoded_resolution = Object.encode_output_resolution(texture.resolution)
+
+    %RustlerFriendly{
```
Janix4000, line 71:
```suggestion
    %__MODULE__{
```

native/membrane_videocompositor/src/elixir_bridge/scene_validation.rs
```diff
@@ -0,0 +1,456 @@
...
+
+        for (_, object) in objects {
+            if let Object::Video(InputVideo { input_pad }) = object {
+                if input_pads.contains(input_pad) {
+                    return Err(SceneParsingError::DuplicatePadReferences(input_pad.clone()));
+                }
+
+                input_pads.insert(input_pad);
+            }
+        }
```
Janix4000, line 69: I believe you can use a more universal function to find duplicates, of smth like this:
```rust
fn fisrt_duplicate_element<T>(iter: T) -> Option<T>
where
    T: IntoIterator,
    T::Item: Eq + Hash,
{
    let mut uniq = HashSet::new();
    iter.into_iter().find(move |x| !uniq.insert(x))
}
```
Based on this [SO answer](https://stackoverflow.com/a/46767732).

native/membrane_videocompositor/src/elixir_bridge/scene_validation.rs
```diff
@@ -0,0 +1,456 @@
...
+        let unused_names = defined_names.difference(&used_names).collect::<Vec<_>>();
+
+        if !unused_names.is_empty() {
+            return Err(SceneParsingError::UnusedObject((*unused_names[0]).clone()));
+        }
+
+        Ok(())
+    }
+
+    /// returns true if a cycle exists in the scene graph
```
Janix4000, line 128: It no longer returns a boolean value

native/membrane_videocompositor/src/elixir_bridge/scene_validation.rs
```diff
@@ -0,0 +1,456 @@
...
+
+        Self::check_for_duplicate_or_undefined_names(&self.objects, &self.output)?;
+
+        Self::check_for_unused_objects(&self.objects, &self.output)?;
+
+        let objects = self
+            .objects
+            .iter()
+            .map(|(name, object)| (name, object))
+            .collect::<HashMap<_, _>>();
```
Janix4000, line 176: [NIT] This collecting procedure is on a different level of abstraction compared to the rest of the body of this function.
I suggest extracting it to the separate helper function.

test/support/mock/transformations/rotate.ex
Janix4000: It is important to stick to our local culture and not import corrupt values and memes from the West.

Janix4000 (changes requested):
Minor code improvements.

lib/membrane/video_compositor/scene/object/texture.ex
jerzywilczek: `RustlerFriendly` is not this module :(

native/membrane_videocompositor/src/elixir_bridge/scene_validation.rs
jerzywilczek: cool!

lib/membrane/video_compositor/scene/object/texture.ex
Janix4000: My bad

Janix4000 (approved):
🥕

jerzywilczek:
The idea of having a separate `impl Scene` in the validation module didn't sit right with me. I decided to move the validation into a new struct called `SceneValidator`. After doing this I also moved the functions that extract `mentioned_names` back into the `elixir_structs` module so that they're implemented next to the definitions of `Object`, `Texture` etc. Please @Janix4000 and @WojciechBarczynski  can you check it again?

native/membrane_videocompositor/src/elixir_bridge/scene_validation.rs
```diff
@@ -122,10 +90,7 @@ impl Scene {
     }
 
     /// returns `Err(SceneParsingError::CycleDetected)` if a cycle exists in the scene graph, `Ok(())` otherwise.
-    fn contains_cycle(
-        objects: &HashMap<&ObjectName, &Object>,
-        final_object: &ObjectName,
-    ) -> Result<(), SceneParsingError> {
+    fn contains_cycle(&self) -> Result<(), SceneParsingError> {
```
Janix4000, line 93:
```suggestion
    fn check_for_cycles(&self) -> Result<(), SceneParsingError> {
```

native/membrane_videocompositor/src/elixir_bridge/scene_validation.rs
```diff
@@ -0,0 +1,411 @@
...
+        let unused_names = defined_names.difference(&used_names).collect::<Vec<_>>();
+
+        if !unused_names.is_empty() {
+            return Err(SceneParsingError::UnusedObject((*unused_names[0]).clone()));
+        }
+
+        Ok(())
+    }
+
+    /// returns `Err(SceneParsingError::CycleDetected)` if a cycle exists in the scene graph, `Ok(())` otherwise.
```
WojciechBarczynski, line 92: [NIT] This is quite obvious from the function name spec. We should either remove this doc or add docs like this to all those functions for consistency.

native/membrane_videocompositor/src/elixir_bridge/elixir_structs.rs
```diff
@@ -308,7 +351,7 @@ impl TryInto<crate::scene::Scene> for Scene {
     type Error = SceneParsingError;
 
     fn try_into(self) -> Result<crate::scene::Scene, Self::Error> {
-        self.validate()?;
+        SceneValidator::new(&self).validate()?;
```
Janix4000, line 354: What is the benefit of using a short-living struct instead of a single function operating on the scene directly?

Janix4000 (changes requested):
Small adjustments

WojciechBarczynski (approved):
LGTM, though I think it would be better for git history clarity to merge #104 before this and add here `StaticFrame`/`Image`

native/membrane_videocompositor/src/elixir_bridge/elixir_structs.rs
jerzywilczek: The `Scene::validate` function was defined in a `impl Scene` in a different module, which is quite unconventional and causes some unintuitive encapsulation behaviour. Since the struct only contains references to the `Scene`, it most likely compiles to the same code.

We could also just define a `validate(scene: &Scene)` function in the validation module, but I somehow like this solution better (the `SceneValidator` sort of 'prepares' the common parameters for the validation functions, which means they don't need to be passed in explicitly).

Janix4000: `SceneValidator` indicates some OO actions and other versions o validations. 

I do not like this pattern, but it is not crucial. I will leave it up to you.

Janix4000 (approved):
K

test/scene_deserialization_hard_test.exs
```diff
@@ -0,0 +1,53 @@
...
+            {:input, 2} => :ball
+          },
+          resolution: %Resolution{width: 1920, height: 1080}
+      }
+    ],
+    output: :final_object
+  }
+
+  test "deserialize" do
+    assert :ok = Native.test_scene_deserialization(Scene.encode(@scene))
```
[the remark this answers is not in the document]
jerzywilczek: maybe when we have a more complete system

lib/membrane/video_compositor/scene/object/layout.ex
```diff
@@ -37,4 +54,80 @@ defmodule Membrane.VideoCompositor.Scene.Object.Layout do
...
+  for now.
+  """
+  @callback encode(t()) :: rust_representation()
+
+  @doc false
+  # Encode the layout to a Layout.RustlerFriendly in order to prepare it for
+  # the rust conversion.
+  @spec encode(t()) :: RustlerFriendly.t()
+  def encode(layout = %module{inputs: inputs, resolution: resolution}) do
+    rust_representation = module.encode(layout)
```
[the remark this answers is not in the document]
jerzywilczek: As we discussed in person, this is rather unnecessary at this stage. perhaps it will be beneficial in the future, and we can implement it then. Right now it is rather pointless.

varsill (approved):

WojciechBarczynski (approved):
LGTM

lib/membrane/video_compositor/object/layout.ex
```diff
@@ -58,9 +58,9 @@ defmodule Membrane.VideoCompositor.Object.Layout do
   defmodule RustlerFriendly do
     @moduledoc false
     # A rustler-friendly version of the Layout, prepared for rust serialization
-    alias Membrane.VideoCompositor.Scene.Object.Layout
-    alias Membrane.VideoCompositor.Scene.Object.RustlerFriendly, as: RFObject
-    alias Membrane.VideoCompositor.Scene.Resolution
+    alias Membrane.VideoCompositor.Object.Layout
+    alias Membrane.VideoCompositor.Object.RustlerFriendly, as: RFObject
+    alias Membrane.VideoCompositor.Resolution
```
varsill: `Object.RustlerFriendly` is on the same level as `Object.Layout` and I am not the biggest fan of it, but I guess there is no way to mitigate it with our approach. In case you can think of some solution, I'm leaving it for a reconsideration :D

varsill (approved):

Janix4000 (changes requested):

test/scene_deserialization_hard_test.exs
```diff
@@ -16,11 +16,47 @@ defmodule Membrane.VideoCompositor.Test.DeserializeHard do
...
+        | z_value: 1.0
+      }
+    },
+    inputs: %{},
+    resolution: nil
+  }
+
+  # binary with placeholder data, matching size of
+  # a single yuv420p full hd frame
+  @background_frame_data <<0::3_110_400>>
```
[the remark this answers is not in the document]
jerzywilczek: No I can't
```ex
  @background_byte_size 1920 * 1080 * 3 / 2
  @background_frame_data <<0::@background_byte_size>>
```
results in:
```
== Compilation error in file test/scene_deserialization_hard_test.exs ==
** (CompileError) test/scene_deserialization_hard_test.exs:49: unknown bitstring specifier: Module.__get_attribute__(__MODULE__, :background_byte_size, 49, true)
    (elixir 1.14.3) src/elixir_bitstring.erl:160: :elixir_bitstring.expand_specs/7
    (elixir 1.14.3) src/elixir_bitstring.erl:38: :elixir_bitstring.expand/8
    (elixir 1.14.3) src/elixir_bitstring.erl:26: :elixir_bitstring.expand/5
    (elixir 1.14.3) expanding macro: Kernel.@/1
    test/scene_deserialization_hard_test.exs:49: Membrane.VideoCompositor.Test.DeserializeHard (module)
```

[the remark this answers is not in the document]
jerzywilczek: unless there's some secret way to do it

Janix4000 (approved):
👩‍🎤

PR #101, opened 2023-04-12
Add initial version of callbacks

DominikWolek:
TODO:

- [x] Design doc
- [x] Docummentation
- [x] Rewrite callbacks' contexts into Membrane Core style (macro `__using__`)
- [x] Split into files
- [x] ~~Maybe tests~~ (what to test?)
- [x] Static input


DominikWolek:
Design doc: https://membraneframework.atlassian.net/l/cp/x2hMDeov

WojciechBarczynski:
@DominikWolek change base to `new-compositor`

lib/membrane/video_compositor/handler/callback_ctx.ex
```diff
@@ -0,0 +1,31 @@
...
+
+      fields_names = unquote(Keyword.keys(fields))
+      default_fields_names = unquote(Keyword.keys(default_fields))
+
+      @enforce_keys Module.get_attribute(__MODULE__, :enforce_keys, fields_names) ++
+                      default_fields_names
+
+      defstruct fields_names ++ default_fields_names
+    end
+  end
```
WojciechBarczynski, line 30: With this macro docs lack `@typedoc`, which IMO would be useful for VC users encountering those `callbacks`. Repeating this logic in every `CallbackCtx.sth` is certainly bad, but I'm curious whether we can somehow generate them, with common parts described here and callback-specific ones (like `earliest_start`) in  `CallbackCtx.sth`.
`default_fields` are rather self-explanatory, but concerted whether it always the case with custom ones.

@DominikWolek What do you think?

lib/membrane/video_compositor/handler/callback_ctx.ex
```diff
@@ -0,0 +1,31 @@
+defmodule Membrane.VideoCompositor.Handler.CallbackCtx do
+  @moduledoc """
+  Structure representing a common part of the context
+  for all of the callbacks.
+  """
+  alias Membrane.{Pad, Time}
+  alias Membrane.VideoCompositor.Scene
+
+  defmacro __using__(fields) do
```
WojciechBarczynski, line 9: [NIT] This seems more explanatory. `fields` suggests, that those are all struct fields, which isn't the case here.
```suggestion
  defmacro __using__(custom_fields) do
```

WojciechBarczynski (changes requested):
1. Change base
2. Without the design doc `earliest_start` might be confusing

lib/membrane/video_compositor/handler/callback_ctx.ex
```diff
@@ -0,0 +1,31 @@
...
+    quote do
+      @type t :: %__MODULE__{unquote_splicing(fields ++ default_fields)}
+
+      fields_names = unquote(Keyword.keys(fields))
+      default_fields_names = unquote(Keyword.keys(default_fields))
+
+      @enforce_keys Module.get_attribute(__MODULE__, :enforce_keys, fields_names) ++
+                      default_fields_names
+
+      defstruct fields_names ++ default_fields_names
```
varsill, line 28: Here you are taking advantage that the quoted representation of the list is a list itself. And as far as I can see it won't allow to define custom default values:
```
defmodule X do
 use CallbackCtx, [x: 1, y: 2]
end

struct =.%X{input_pads: ..., scenes_queue: ..., current_scene: ...} # error
```

varsill (changes requested):

lib/membrane/video_compositor/handler/callback_ctx/video_add.ex
```diff
@@ -0,0 +1,8 @@
+defmodule Membrane.VideoCompositor.Handler.CallbackCtx.VideoAdd do
```
Janix4000, line 1: Isn't `VideoAdded` more sufficient?

Janix4000 (changes requested):

lib/membrane/video_compositor/handler/callback_ctx.ex
DominikWolek: IMO we can simply write moduledocs for callback contexts

lib/membrane/video_compositor/handler/callback_ctx.ex
```diff
@@ -0,0 +1,31 @@
...
+      quote do
+        [
+          input_pads: list(Pad.ref_t()),
+          scenes_queue: list({scene :: Scene.t(), start_scene_timestamp :: Time.t()}),
+          current_scene: Scene.t()
+        ]
+      end
+
+    quote do
+      @type t :: %__MODULE__{unquote_splicing(fields ++ default_fields)}
```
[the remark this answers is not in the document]
DominikWolek: We could add it, but IMO it's not needed and mistakes like that will be catched in code review.

lib/membrane/video_compositor/handler/callback_ctx/video_add.ex
DominikWolek: No, cause that implies past.

lib/membrane/video_compositor/handler/callback_ctx.ex
varsill: Oh, apparently I have got it all wrong :D The "values" passed in the keyword list are about to be type specification, not the default values of the resulting keyword list

lib/membrane/video_compositor/handler.ex
```diff
@@ -0,0 +1,75 @@
...
+
+  @type video_details :: {pad :: Pad.ref_t(), format :: StreamFormat.t()}
+
+  @typedoc """
+  Type of a valid return value from a callback.
+  """
+  @type callback_return :: immediate_callback_return() | timed_callback_return()
+
+  @typedoc """
+  Type of a valid return value from most of the callbacks.
```
varsill, line 26: I would still specify, that in this case the scene is changed immediately, rather than saying that it's common for most callbacks

mix.exs
```diff
@@ -1,7 +1,7 @@
 defmodule Membrane.VideoCompositor.Mixfile do
   use Mix.Project
 
-  @version "0.3.0"
+  @version "0.3.1"
```
varsill, line 4: Are we sure we want to bump the version?

lib/membrane/video_compositor/handler/callback_ctx.ex
varsill: With the use of macros we can append the "common" part of the moduledoc to the user-defined moduledoc. 
In Core we do it i.e. here: https://github.com/membraneframework/membrane_core/blob/master/lib/membrane/core/docs_helper.ex
I believe we can also implement that strategy in this case.

lib/membrane/video_compositor/handler.ex
```diff
@@ -0,0 +1,56 @@
...
+              ctx :: CallbackCtx.VideoRemove.t(),
+              state :: state()
+            ) :: callback_return()
+
+  @doc """
+  Callback invoked when video compositor receives a message
+  that is not recognized as an internal membrane message.
+  """
+  @callback handle_info(msg :: any(), ctx :: CallbackCtx.Info.t(), state :: state()) ::
+              callback_return() | {start_ts :: Time.t(), callback_return()}
```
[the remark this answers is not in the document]
WojciechBarczynski: It's fine now. We can add here a link to the design doc once we publish them.

lib/membrane/video_compositor/handler.ex
```diff
@@ -0,0 +1,75 @@
+defmodule Membrane.VideoCompositor.Handler do
+  @moduledoc """
+  Module defining behaviour of handlers.
+
+  Implementing handler allows to provide custom implementation and
+  react to various events, among others by setting a new scene and/or state.
```
WojciechBarczynski, line 6: From this description, someone might misunderstand the `setting state` phrase.

There are a lot of `states` in the video compositor and I'm not sure whether it's clear, that this is only a custom state passed to the callback, that is only set by the user implementing callbacks and isn't used in any part of VC logic at all. Moreover, combining it in one alternative/conjunction with the `scene` specifying VC composition logic, suggests that this state is somehow used by the VC backend. 

I think that some indication that VC doesn't interact with this state at all is needed.

mix.exs
DominikWolek: This was merged from the master

varsill (approved):

lib/membrane/video_compositor/handler/callback_ctx.ex
```diff
@@ -0,0 +1,29 @@
...
+  """
+  alias Membrane.{Pad, Time}
+  alias Membrane.VideoCompositor.Scene
+
+  defmacro __using__(custom_fields) do
+    default_fields =
+      quote do
+        [
+          input_pads: list(Pad.ref_t()),
+          scenes_queue: list({scene :: Scene.t(), start_scene_timestamp :: Time.t()}),
```
WojciechBarczynski, line 14: [NIT] Usually in priority queues, the priority level is the first argument. `scenes_queue` is just the priority queue with the lowest `start_scene_timestamp` and `scenes` having the highest priority. 
```suggestion
          scenes_queue: list({start_scene_timestamp :: Time.t(), scene :: Scene.t()}),
```

lib/membrane/video_compositor/handler/callback_ctx.ex
```diff
@@ -0,0 +1,29 @@
...
+  for all of the callbacks.
+  """
+  alias Membrane.{Pad, Time}
+  alias Membrane.VideoCompositor.Scene
+
+  defmacro __using__(custom_fields) do
+    default_fields =
+      quote do
+        [
+          input_pads: list(Pad.ref_t()),
```
WojciechBarczynski, line 13: `input_pads` is a little bit ambiguous when it comes to `handle_init`, `handle_video_add`, and `handle_video_remove` callbacks. 

IMO there isn't one obvious answer to those questions:
1.  Does `input_pads` in `Membrane.VideoCompositor.Handler.CallbackContext.VideoAdd` contains just added video?
2. Does `input_pads` in `Membrane.VideoCompositor.Handler.CallbackContext.VideoRemove` contains the just removed video?
3. What it contains in `Membrane.VideoCompositor.Handler.CallbackContext.Init`?

So many questions, so few answers...

Maybe leaving just a note somewhere in the docs, that this should e.x. contain

WojciechBarczynski (changes requested):

lib/membrane/video_compositor/handler.ex
```diff
@@ -0,0 +1,77 @@
...
+  sending custom messages and react to them. Therefore, it allows to
+  react to custom events not specified by the other callbacks.
+
+  Please be minded that this callback does not have fixed start
+  and you can specify start by returning `t:timed_callback_return/0`.
+  The earliest start is specified in `:earliest_start` field of
+  `Membrane.VideoCompositor.Handler.CallbackContext.Info`.
+  """
+  @callback handle_info(msg :: any(), ctx :: CallbackContext.Info.t(), state :: state()) ::
+              callback_return()
```
WojciechBarczynski, line 76: [NIT] Calling type `callback_return` suggests, that it's the return of all callbacks, rather than just a return of `handle_info` callback.

```suggestion
  @callback handle_info(msg :: any(), ctx :: CallbackContext.Info.t(), state :: state()) ::
              immediate_callback_return() | timed_callback_return()
```

lib/membrane/video_compositor/handler.ex
```diff
@@ -0,0 +1,77 @@
...
+  Type of user-managed inner state of the handler.
+  """
+  @type state :: any()
+
+  @type video_details :: {pad :: Pad.ref_t(), format :: StreamFormat.t()}
+
+  @typedoc """
+  Type of a valid return value from a callback.
+  """
+  @type callback_return :: immediate_callback_return() | timed_callback_return()
```
WojciechBarczynski, line 24:
```suggestion
```

lib/membrane/video_compositor/compositor_element.ex
```diff
@@ -110,6 +110,8 @@ defmodule Membrane.VideoCompositor.CompositorElement do
 
   @impl true
   def handle_pad_added(pad, context, state = %State{}) do
+    # IO.inspect(callback: "addded", pad: pad, context: context.pads)
+
```
WojciechBarczynski, line 114:
```suggestion
```

lib/membrane/video_compositor/handler.ex
```diff
@@ -0,0 +1,79 @@
...
+
+  @typedoc """
+  Type of a valid return value from callback not changing the current scene.
+  """
+  @type idle_callback_return :: {state :: state()}
+
+  @doc """
+  Callback invoked upon initialization of Video Compositor.
+  """
+  @callback handle_init(ctx :: CallbackContext.Init.t()) :: immediate_callback_return()
```
WojciechBarczynski, line 40: IMO this ctx is quite misleading. Maybe VC options or not ctx at all would be better.

Screenshot 2023-05-05 at 12 31 09

lib/membrane/video_compositor/handler/callback_ctx.ex
```diff
@@ -0,0 +1,58 @@
+defmodule Membrane.VideoCompositor.Handler.CallbackContext do
+  @moduledoc """
+  Structure representing a common part of the context
+  for all of the callbacks.
+  """
+  alias Membrane.{Pad, Time}
+  alias Membrane.VideoCompositor.Scene
+
+  @spec __using__(any) ::
+          {:__block__, [], [{:= | :@ | :defstruct | {any, any, any}, [...], [...]}, ...]}
```
WojciechBarczynski, line 10:
```suggestion
```

WojciechBarczynski (changes requested):

lib/membrane/video_compositor/handler.ex
WojciechBarczynski: `current_scene` should be nil or empty, `input_pads` and `scenes_queues` are just empty lists here, so I think we don't need `ctx` here.

WojciechBarczynski (approved):
LGTM

PR #102, opened 2023-04-13
Fix wgpu interpolation (`filter mode`)

WojciechBarczynski:
This PR fix wgpu `filter mode` setup. We change it (interpolation algorithm) from `Nearest` to `Linear`, which is more suitable for majority of uses cases. Maybe later on, we'll add something like an optional `wgpu_setup` VC init option, providing users with the possiblity to easly customize such options.

I had to change some tests and refs, to handle new interpolation algorithm.


DominikWolek (commented):
Overall nice quick fix

DominikWolek (approved):

PR #104, opened 2023-04-17
Design scene static inputs and flatten VC modules structure

WojciechBarczynski:
This PR adds `InputImage` to VideoCompositor and flattens the VC module structure.

Static Inputs are useful when the user wants to use input, that doesn't change between frames (e.g. some background). Without static inputs, users have to artificially convert a single frame into a video stream, which this PR aims to simplify.

Aside from adding `InputImage`, I modified the hard example to showcase API.

To flatten the project structure, I moved all modules nested in the `scene` struct to parent modules, such that the project structure isn't that deeply nested. With that, the docs should be a little bit more readable.


lib/membrane/video_compositor/scene/object/input/static_frame.ex
```diff
@@ -0,0 +1,26 @@
+defmodule Membrane.VideoCompositor.Scene.Object.Input.StaticFrame do
```
jerzywilczek, line 1: A 'Static Frame' is just an image. I'm not 100% sure, but I would consider renaming this module to `Image`.

lib/membrane/video_compositor/examples/mock/assets/membrane_logo.raw
jerzywilczek, line 1: 1.48MB for a logo? That's quite a lot...

lib/membrane/video_compositor/scene/object/input/static_frame.ex
jerzywilczek, line 1: I'm also not 100% sure about having `Input` in the module path (or whatever that's called in elixir :roll_eyes:) and creating a separate directory just for `StaticFrame` and `Video`. It seems to put more cognitive load on the person reading the code with not much benefit (since we're only moving two files to the separate folder)

jerzywilczek (approved):
The PR is generally ready to merge, I just have a couple of small suggestions and worries.

lib/membrane/video_compositor/scene/object/input/static_frame.ex
WojciechBarczynski: @DominikWolek What do you think?

DominikWolek: It's called module name. I agree, since we don't have `scene/object/input.ex` file

I'd also flatten other parts of the structure, e.g. `.Scene.Object`to `.Object`

lib/membrane/video_compositor/examples/mock/assets/membrane_logo.raw
DominikWolek: It's a nice logo :D But I agree

lib/membrane/video_compositor/scene/object/input/static_frame.ex
```diff
@@ -0,0 +1,26 @@
...
+  @moduledoc """
+  Structure representing `StaticFrame` wrapper for using single frame as
+  an object in a scene. It's useful, when user want to use some single image
```
DominikWolek, line 4:
```suggestion
  Structure representing `StaticFrame` wrapper for using single frame as
  an object in a scene. 
  
  It's useful, when user want to use some single image
```

lib/membrane/video_compositor/scene/object/input/static_frame.ex
```diff
@@ -0,0 +1,26 @@
...
+  """
+
+  alias Membrane.RawVideo
+
+  @enforce_keys [:frame, :stream_format]
+  defstruct @enforce_keys
+
+  @typedoc """
+  Defines wrapper for static VC frame InputStatic.
+  The `frame` is raw video binary.
```
DominikWolek, line 16:
```suggestion
  @typedoc """
  Defines wrapper for static VC frame InputStatic.
  
  The `frame` is raw video binary.
```

DominikWolek (approved):

lib/membrane/video_compositor/scene/object/input/static_frame.ex
jerzywilczek:
```suggestion
  @typedoc """
  Defines wrapper for static VC frame InputStatic.
  
  The `frame` is a raw video binary.
```

lib/membrane/video_compositor/examples/mock/assets/membrane_logo.raw
jerzywilczek: i think that would be fine

lib/membrane/video_compositor/scene/object/input/static_frame.ex
WojciechBarczynski: I changed the project module's structure. It's quite a major change, so can you check it out and let me know whether you like it? @DominikWolek @jerzywilczek

lib/membrane/video_compositor/object/input_image.ex
```diff
@@ -1,7 +1,9 @@
-defmodule Membrane.VideoCompositor.Scene.Object.Input.StaticFrame do
+defmodule Membrane.VideoCompositor.Object.InputImage do
```
jerzywilczek, line 1: I think I like `Image` and `Video` more than `InputImage` and `InputVideo`.
What do you think?

jerzywilczek (changes requested):

lib/membrane/video_compositor/object/input_image.ex
jerzywilczek: I actually think it's fine as it is

jerzywilczek (approved):
:rocket:

PR #106, opened 2023-05-08
Ms 542 transfer

jerzywilczek:
* Create a new crate (membrane_video_compositor_common) that is
   designed to hold the components necessary for communication between
   custom plugins and the compositor.

 * Add some simple structures to this crate. The structures are in a
   very basic form for now and will be changed when the proper plugin
   API will be designed and added.

 * Move the dependency on `wgpu` to this new crate and republish it so
   the compositor and plugins are using the same wgpu version.

 * Put both crates (the new one and the compositor) in a cargo workspace
   and refactor the tests to run on all crates in the workspace.

 * Update rustler for good measure.
 
 * Rename the compositor crate from `membrane_videocompositor` to `membrane_video_compositor` for consistency.


native/membrane_video_compositor_common/src/elixir_transfer.rs
```diff
@@ -0,0 +1,108 @@
...
+impl<'a> rustler::Decoder<'a> for ElixirTypePacket {
+    fn decode(term: rustler::Term<'a>) -> rustler::NifResult<Self> {
+        let pointer: (usize, usize) = rustler::Decoder::decode(term)?;
+        Ok(Self { pointer })
+    }
+}
+
+impl rustler::Encoder for ElixirTypePacket {
+    fn encode<'a>(&self, env: rustler::Env<'a>) -> rustler::Term<'a> {
+        self.pointer.encode(env)
```
varsill, line 38: TBH I have some trust issues with that one :DD 
How is that supposed to work, when `pointer` is `(usize, usize)`?

native/membrane_video_compositor_common/src/elixir_transfer.rs
varsill, line 1: Both the struct's methods implementations and the traits implementations seem to be quite repeatable. Is it supposed to change when we continue work on rust-elixir-rust bridge?

varsill (commented):
Looks good to me, I've just got some tiny doubts I prefer to clarify before approve'ing ;)

native/membrane_video_compositor_common/src/elixir_transfer.rs
jerzywilczek: the `rustler::Encoder` trait is automatically implemented by rustler for all tuples with lengths <= 6 provided that the types of all members of those tuples also implement `rustler::Encoder`. Since `usize` implements `rustler::Encoder`, tuples containing `usize`s also do.

mix.exs
```diff
@@ -83,7 +83,7 @@ defmodule Membrane.VideoCompositor.Mixfile do
         ["lib", "mix.exs", "README*", "LICENSE*", ".formatter.exs"] ++
           Enum.map(
             ["src", ".cargo/config", "Cargo.toml", "Cargo.lock"],
-            &"native/membrane_videocompositor/#{&1}"
+            &"native/membrane_video_compositor/#{&1}"
```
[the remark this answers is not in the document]
jerzywilczek: I actually am not sure what this field is for

native/membrane_video_compositor_common/src/elixir_transfer.rs
jerzywilczek: We have three different 'classes' of things that we want to transfer through elixir:
 - custom structs defined by the plugin
 - structs that implement a specific trait (Transformation or Layout)
 - structs which are used both in the compositor and the plugins
 
The first one is currently called `ElixirTypePacket`, I want to change the name today.
The last one is not implemented in this PR currently, I only realized the necessity for such a struct today.
The worst one for repetition is the second one, since it requires a separate implementation for each trait. What would be ideal is if rust allowed us to be generic over a trait, but this is sadly not possible yet.

I believe we need at least one implementation for each of the three and that this is rather reasonable. The most unfortunate part is the doubled implementation for each of the traits, I think it would be possible to implement this with a macro.

So, the conclusion is:
I propose to:
Keep `ElixirTypePacket` as-is (but change it's name)
Add a new struct for types that are known when compiling the compositor (the ones that need to be modified by both the client and the compositor)
Implement `ElixirLayoutPacket` and `ElixirTransformationPacket` with a macro, which allows us to keep a single implementation.

Do you think this is a good solution?

native/membrane_video_compositor_common/src/elixir_transfer.rs
```diff
@@ -0,0 +1,108 @@
...
+        let pointer: (usize, usize) = unsafe { std::mem::transmute(payload) };
+
+        ElixirTypePacket { pointer }
+    }
+
+    /// # Safety
+    /// The caller must ensure that all of the following conditions are met:
+    ///  - The packet must have been encoded in the same address space as the call site
+    ///  - The packet must have been encoded in a crate compiled in the same compilation as the call site
+    ///  - The packet must not have been cloned or duplicated in any way in elixir
```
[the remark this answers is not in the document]
varsill: I get it. At first I thought that the problem would be caused by just the process of copying of the data on the Elixir side (like it was about to get corrupted along that process) and it seemed quite exotic to me. You can slightly rephrase it and mention, that the caller must ensure that there is no other `decode` call on the same packet, but I leave it for you to decide.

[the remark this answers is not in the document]
jerzywilczek: Makes sense!

mix.exs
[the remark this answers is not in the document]
varsill: I believe it's used to describe which files should be put in a package

native/membrane_video_compositor_common/src/elixir_transfer.rs
varsill: Good idea, the macro, (similar to `derive`)  for these second class types should solve the problem ;)

native/membrane_video_compositor_common/src/elixir_transfer.rs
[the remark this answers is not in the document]
jerzywilczek: Changed the docs, please let me know if it's easier to understand now

native/membrane_video_compositor_common/src/elixir_transfer.rs
jerzywilczek: Done, please check whether the implementation looks good

mix.exs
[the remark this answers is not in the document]
jerzywilczek: I don't really know whether to put it here, since it depends on how we want to distribute `membrane_video_compositor_common`. I guess I will put it here for now and we can think about this in a more opportune time.

native/membrane_video_compositor_common/src/elixir_transfer.rs
```diff
@@ -0,0 +1,130 @@
+use std::{any::Any, marker::PhantomData, sync::Arc};
+
+use crate::plugins::{layout::UntypedLayout, transformation::UntypedTransformation};
+
+pub struct ElixirCustomStructPacket {
```
varsill, line 5:
```suggestion
pub struct CustomStructElixirPacket {
```

varsill (approved):
lgtm 🥇

PR #108, opened 2023-05-12
Tag nifs properly (as dirty IO-bound)

jerzywilczek:
We have been doing this wrong for a very long time. This was caused by rustler's poor documentation of this feature.


DominikWolek (approved):
No idea how we missed it till now, good catch.

WojciechBarczynski (approved):

varsill (approved):

PR #110, opened 2023-06-05
Ms 530 initialization

jerzywilczek:
Initialization is performed as follows:
 * when initializing the compositor, you need to pass the
   transformations and layouts (elixir modules) you want to use.
 * then, all of those modules get their `initialize/1` function called.
   This function creates the initialized rust versions of those
   transformations and returns them.
 * those rust versions then get registered in the compositor.


native/membrane_video_compositor_common/src/plugins/transformation.rs
```diff
@@ -0,0 +1,72 @@
+use std::{any::Any, sync::Arc};
+
+use crate::WgpuContext;
+
+// Question to the reviewers: should the example below be in the finished documentation? Should it be
+//                            kept until the end of the plugins PR?
+//
+// Question to the reviewers, but for later: There are a couple of syntax sugar possibilities for the initialization described below.
+//                                           We could have a proc macro, a macro_rules macro, or possibly even leave it as-is but
+//                                           generate the boilerplate. Which one should we use?
```
varsill, line 10: I would say that `macro_rules` macro would be sufficient so that not to repeat the implementations

native/membrane_video_compositor_common/src/plugins/transformation.rs
```diff
@@ -0,0 +1,72 @@
+use std::{any::Any, sync::Arc};
+
+use crate::WgpuContext;
+
+// Question to the reviewers: should the example below be in the finished documentation? Should it be
+//                            kept until the end of the plugins PR?
```
varsill, line 6: I would definitely leave it here right now, once the rest of the system is complete we can either put it in the documentation, or at least use it somewhere in tests (but I would opt for the first one)

varsill (changes requested):

native/membrane_video_compositor_common/src/plugins/transformation.rs
jerzywilczek: we'll need to discuss this a bit later, when it's time for introducing sugaring. I was more on the side of generating the boilerplate -- a mix task like `mix compositor.gen transformation`, which would generate a file with two functions you need to fill in yourself. Don't know if this would be possible yet though.

lib/membrane/video_compositor/compositor_element.ex
```diff
@@ -38,16 +40,18 @@ defmodule Membrane.VideoCompositor.CompositorElement do
 
     @type t() :: %__MODULE__{
             wgpu_state: wgpu_state_t(),
+            new_compositor_state: WgpuAdapter.new_compositor_state(),
```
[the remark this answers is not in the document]
jerzywilczek: I removed all of the old compositor and only left the new stuff. Also renamed this state to `native_state`

native/membrane_video_compositor_common/src/plugins/transformation.rs
jerzywilczek: Please leave this unresolved to allow other reviewers to comment on this

lib/membrane/video_compositor/wgpu/native.ex
```diff
@@ -4,57 +4,38 @@ defmodule Membrane.VideoCompositor.Wgpu.Native do
...
-  @spec update_transformations(wgpu_state(), id(), VideoTransformations.t()) ::
-          :ok | {:error, error()}
-  def update_transformations(_state, _id, _transformations), do: error()
+  @spec init() :: {:ok, native_state()} | {:error, error()}
+  def init(), do: error()
 
-  @spec remove_video(wgpu_state(), id()) ::
-          :ok | {:error, error()}
-  def remove_video(_state, _id), do: error()
+  @spec wgpu_ctx(native_state()) :: wgpu_ctx()
```
varsill, line 21: Don't we want to add some alternative return type for indicating failure?

lib/membrane/video_compositor/wgpu/native.ex
```diff
@@ -4,57 +4,38 @@ defmodule Membrane.VideoCompositor.Wgpu.Native do
...
+          native_state(),
+          Membrane.VideoCompositor.Object.Layout.initialized_layout()
+        ) :: :ok | {:error, error()}
+  def register_layout(_state, _layout), do: error()
+
+  @spec mock_transformation(wgpu_ctx()) ::
+          Membrane.VideoCompositor.Transformation.initialized_transformation()
+  def mock_transformation(_wgpu_ctx), do: error()
 
   defp error(), do: :erlang.nif_error(:nif_not_loaded)
```
varsill, line 40: `:erlang.nif_error` does not follow the specification of return types of functions, in which it is used, but surely it's not important right now.

lib/membrane/video_compositor/wgpu/wgpu_adapter.ex
```diff
@@ -1,165 +1,79 @@
...
-
-    case Native.add_video(state, id, rust_stream_format, placement, transformations) do
-      :ok -> :ok
-      {:error, reason} -> raise "Error while adding a video, reason: #{inspect(reason)}"
-    end
+  @spec register_transformations(
+          native_state(),
+          list(Transformation.transformation_module())
+        ) :: :ok
+  def register_transformations(state, transformations) do
```
varsill, line 46: perhaps `add_transformations` or something similar would be better than `register_transformations` since the function not only "registers" the transformations in the native part, but it also initializes them?

varsill (commented):
Just a few more things consider

lib/membrane/video_compositor/wgpu/native.ex
jerzywilczek: This can't fail

jerzywilczek: This only increments the refcount of a shared pointer and gives you the pointer.

lib/membrane/video_compositor/wgpu/wgpu_adapter.ex
jerzywilczek: That's very fair

varsill (approved):
🥇

native/membrane_video_compositor_common/src/plugins/transformation.rs
WojciechBarczynski: IMO we can leave it here

native/membrane_video_compositor_common/src/plugins/transformation.rs
WojciechBarczynski: If VC will be used only with Elixir, I would rather opt for creating a mix task. 
If we plan to integrate it with sth else in the future, the macro is probably a better option. 

I think this is outside of the scope of this task/PR.

native/membrane_video_compositor/src/elixir_bridge/elixir_structs.rs
```diff
@@ -466,7 +337,7 @@ mod tests {
                     Object::Layout(Layout {
                         inputs,
                         resolution: LayoutOutputResolution::Name(a),
-                        implementation: 42,
+                        implementation: (42, 42),
```
WojciechBarczynski, line 340: This is a little inconsistent. In the `Object::Layout` reference to a specific implementation of layouts is called `implementation`, while in the `Object::Texture` list of the same type, references (references to plugin-provided implementations of Layout/Transformation) are called `transformations`.

Maybe we should normalize that.

WojciechBarczynski (changes requested):
Overall, the core logic looks good. I left a few minor remarks.

lib/membrane/video_compositor/rust_structs/base_video_placement.ex
```diff
@@ -1,33 +0,0 @@
...
-
-  @type t :: %__MODULE__{
-          position: {integer(), integer()},
-          size: {non_neg_integer(), non_neg_integer()},
-          z_value: float()
-        }
-
-  @enforce_keys [:position, :size]
-  defstruct [:position, :size, z_value: 0.0]
-end
```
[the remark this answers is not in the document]
WojciechBarczynski: Obviously, we should leave:
```elixir
      nest_modules_by_prefix: [
        Membrane.VideoCompositor
      ]
```

lib/membrane/video_compositor/wgpu/native.ex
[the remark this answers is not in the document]
jerzywilczek: why `Spec` in the first case?

maybe `Membrane.VideoCompositor.Wgpu.Native` -> `Membrane.VideoCompositor.Native.Impl`?

lib/membrane/video_compositor/wgpu/wgpu_adapter.ex
```diff
@@ -1,165 +1,79 @@
 defmodule Membrane.VideoCompositor.WgpuAdapter do
   @moduledoc false
 
-  alias Membrane.VideoCompositor.RustStructs
-  alias Membrane.VideoCompositor.VideoTransformations
+  alias Membrane.VideoCompositor.Object.Layout
+  alias Membrane.VideoCompositor.Transformation
```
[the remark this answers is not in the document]
jerzywilczek: the problem is that `Layout` is on one hand on the same abstraction layer as `Texture` (it's a node in a scene graph), but on the other on the same abstraction level as `Transformation` (is the set of parameters for a layout implementation and its module implements the behaviour connecting it to a `Layout` trait impl). If we move it like that we will have the same lack of consistency with `Texture` (why is one module that represents a scene object outside of the objects directory?)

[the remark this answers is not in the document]
jerzywilczek: I don't really know how to tackle this, since splitting the current `Layout` into separate modules so that one is a scene node and the other contains the implementation seems unnecessary and just not very good to me

lib/membrane/video_compositor/object/layout.ex
```diff
@@ -86,14 +91,26 @@ defmodule Membrane.VideoCompositor.Object.Layout do
...
   @typedoc """
-  This documentation is temporary, since something needs to be here and it
-  is currently unknown what this will look like.
+  A rust representation of the layout as defined in a scene graph, passed through elixir
+  in an opaque way. In other words, those are the parameters that will be passed to the
+  initialized layout.
+
+  Keep in mind the layout needs to be registered before it's used in a scene graph.
+  """
+  @opaque rust_representation :: {non_neg_integer(), non_neg_integer()}
```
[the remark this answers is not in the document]
jerzywilczek: i changed it to `encoded_params`

native/membrane_video_compositor/src/elixir_bridge/elixir_structs.rs
jerzywilczek: Renamed it to `params`. with this change at least the name is accurate.

lib/membrane/video_compositor/object/layout.ex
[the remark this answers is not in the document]
WojciechBarczynski: Perfect. It's even better.

lib/membrane/video_compositor/wgpu/native.ex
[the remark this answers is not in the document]
WojciechBarczynski: `Membrane.VideoCompositor.Native.Impl` is great :)

lib/membrane/video_compositor/wgpu/wgpu_adapter.ex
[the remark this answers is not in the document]
WojciechBarczynski: What do you think about:
- Scene: `Membrane.VideoCompositor.Scene`
- Object: `Membrane.VideoCompositor.Scene.Object`
- Layout: `Membrane.VideoCompositor.Scene.Layout`
- Texture: `Membrane.VideoCompositor.Scene.Texture`
- Transformation: `Membrane.VideoCompositor.Scene.Transformation`

I know that it would make more sense to have:
- Scene: `Membrane.VideoCompositor.Scene`
- Object: `Membrane.VideoCompositor.Scene.Object`
- Layout: `Membrane.VideoCompositor.Scene.Object.Layout`
- Texture: `Membrane.VideoCompositor.Scene.Object.Texture`
- Transformation: `Membrane.VideoCompositor.Scene.Object.TextureTransformation`

but this level of module nesting is horrible to work with.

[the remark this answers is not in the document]
jerzywilczek: seems ok, do you want me to change it in this PR or in a big renaming and restructuring PR that i'm planning to do soon?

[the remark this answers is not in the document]
WojciechBarczynski: I would change it in this one, since after removing old VC stuff done here, the new project structure makes more sense.

WojciechBarczynski (approved):
GJ 🚀

PR #112, opened 2023-06-07
Design VC callbacks

WojciechBarczynski:
Design VC scene update callbacks


lib/membrane/video_compositor/handler/callback_context/init.ex
```diff
@@ -0,0 +1,19 @@
+defmodule Membrane.VideoCompositor.Handler.CallbackContext.Init do
```
mat-hek, line 1: In the core, we resigned from having separate structs for context. Instead, we have a map and a single module with the type and description of fields common for all callbacks. All callback-specific fields are described in specs and docs of particular callbacks.

lib/membrane/video_compositor/handler.ex
```diff
@@ -0,0 +1,40 @@
...
+  @doc """
+  Callback invoked upon change of VC input videos.
+  Events changing input videos:
+  - video added
+  - video removed
+  - video stream format change
+  """
+  @callback handle_inputs_change(
+              inputs_description :: InputsDescription.t(),
+              ctx :: CallbackContext.InputsChange.t(),
```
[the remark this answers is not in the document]
WojciechBarczynski: Fixed

lib/membrane/video_compositor/handler/callback_context/init.ex
mat-hek: yup, but they may have one day, and a separate struct for each callback seems like overkill anyway

WojciechBarczynski: What do you think about using common `callback_context` for all callbacks with "common context", and using separate ones for callbacks with different contexts? 
Currently, we have `Membrane.VideoCompositor.Handler.CallbackContext` with common callback context and `Membrane.VideoCompositor.Handler.CallbackContext.Init` specific to `handle_init`, since fields in those are disjoint.

mat-hek: My point is to:
- change `Membrane.VideoCompositor.Handler.CallbackContext` to a map
- remove `Membrane.VideoCompositor.Handler.CallbackContext.Init` and describe it in `handle_init` docs
- consider putting the `init_opts` as a separate argument to `handle_init` like it's done everywhere else, then there would be no need for the context in `handle_init`

mat-hek (approved):

lib/membrane/video_compositor/handler.ex
```diff
@@ -0,0 +1,77 @@
...
+  """
+  @callback handle_init(init_options :: VideoCompositor.init_options()) :: state()
+
+  @doc """
+  Callback invoked upon change of VC `t:inputs()`.
+
+  `inputs` changing input videos:
+  - video added
+  - video removed
+  - video stream format change
```
Karolk99, line 57: What about metadata? 
Example: there is an input track that is at first a camera stream and then it changes into a screen share.
Do I have to write custom messages on metadata changes but in this case how to change the context for future calls?

WojciechBarczynski: The `metadata` is passed as an input pad option.
If sth can dynamically change in your application, passing it as a linking option and allowing to modify it, doesn't sound like a good idea. If we allow changing it here and sth crashes, error logs will show different metadata in membrane contexts and VC logs, making it more difficult to debug.

In the case of screen sharing, I would do it like this:
1. Pass the initial "sharing state" option as `metadata` in `pad options`, like this: `via_in(.., options: [..., metadata: %{initial_sharing_state: :camera}]) |> get_child(:video_compostior)`
2. On `handle_inputs_change` check if you have saved the initial option in your `state`. If not, put it in your `state`. Generate a `scene` based on your `state`
3. In case of changing the "sharing state" without changing the input pad, send a message into the VC process and handle it in `handle_info`, where you update the pad "sharing state" in your `state` and generate a `scene`.

```elixir
defmodule VideoCallWithScreenShareHandler do
  @moduledoc false

  @behaviour Membrane.VideoCompositor.Handler

  @impl true
  def handle_init(init_options) do
    %{sharing_state: %{}}
  end

  @impl true
  def handle_inputs_change(inputs, _ctx, state = %{sharing_state: sharing_state}) do
    state = %{
      state
      | sharing_state:
          Map.new(
            inputs,
            fn {pad, pad_properties} ->
              Map.get(sharing_state, pad, pad_properties.metadata.initial_pad_sharing_state)
            end
          )
    }

    {get_scene(state), state}
  end

  @impl true
  def handle_info(msg, ctx, state) do
    case msg do
      {:sharing_state, pad, pad_sharing_state} ->
        state = %{state | sharing_state: Map.put(state.sharing_state, pad, pad_sharing_state)}
        {get_scene(state), state}
      _else ->
        {ctx.current_scene, state}
    end
  end
end
```

Karolk99 (approved):

PR #113, opened 2023-06-07
Ms 434 temporal api

DominikWolek:
This PR covers API for temporal scenes, including:
- scenes expiring after set time
- sequences of expiring scenes
- loops of scenes' sequences
- new handler's callback, `handle_scene_expire` called upon expiration of finite scene

API for temporal transformations will be covered in another PR


lib/membrane/video_compositor/handler.ex
```diff
@@ -59,6 +61,16 @@ defmodule Membrane.VideoCompositor.Handler do
...
+  @doc """
+  Callback invoked upon expiration of Temporal Scene.
+
+  See `Membrane.VideoCompositor.Scene.Temporal`.
+  """
+  @callback handle_scene_expire(
+              ctx :: CallbackContext.SceneExpire.t(),
+              state :: state()
+            ) :: immediate_callback_return()
+
```
WojciechBarczynski, line 73: This may be inconsistent with our callbacks API. In other callbacks, we pass an event related to that callback invocation as a first argument, while here we don't. One may argue, that the `ctx` contains an expired scene, and passing it once more is repetitive and unnecessary, but I think we might solve it. The `ctx` contains
```elixir
current_scene: Membrane.VideoCompositor.Scene.t()
```
meaning we can't pass there
```elixir
Membrane.VideoCompositor.Scene.Temporal.t()
```

I imagine, that sometimes it might be helpful in this callback to get access not only to the last scene but also to the whole expired scene specification (e.g. to the expired `sequence`). Therefore, I would opt for that API:
```suggestion
  @doc """
  Callback invoked upon expiration of Temporal Scene.

  See `Membrane.VideoCompositor.Scene.Temporal`.
  """
  @callback handle_scene_expire(
              expired_scene :: Membrane.VideoCompositor.Scene.Temporal.t(),
              ctx :: CallbackContext.SceneExpire.t(),
              state :: state()
            ) :: immediate_callback_return()

```

But here we run into some ambiguity, that I discovered while writing #112. Typing `CallbackContext.SceneExpire.t()` might suggest, that it holds expired `Membrane.VideoCompositor.Scene.Temporal.t()`, while it doesn't. This is quite problematic since all our callbacks receive `Membrane.VideoCompositor.Handler.CallbackContext.[callback_name].t()`. As we see here and in #112 , this certainly isn't the perfect solution, with the possibility of being misleading.

I suggest that we change our callback API, to the one proposed in #114 .  If implemented in this PR, we should also modify the PR name, since it's certainly outside of the scope of `designing the temporal scene API`.

lib/membrane/video_compositor/scene/temporal.ex
```diff
@@ -0,0 +1,108 @@
+defmodule Membrane.VideoCompositor.Scene.Temporal do
```
WojciechBarczynski, line 1: I think this would be better:
```suggestion
defmodule Membrane.VideoCompositor.TemporalScene do
```

WojciechBarczynski (changes requested):
To avoid going back and forth in this PR, I just created #114 with all of my proposed changes.
Overall I like the design of the temporal scene that we came up with, but I have some suggestions for refining the callbacks system.

lib/membrane/video_compositor/handler.ex
```diff
@@ -22,12 +22,14 @@ defmodule Membrane.VideoCompositor.Handler do
   Type of a valid return value from the callback. By returning this type,
   the scene will be changed immediate, i.e. at the moment when the event happens.
```
bblaszkow06, line 23:
```suggestion
  the scene will be changed immediately, i.e. at the moment when the event happens.
```

WojciechBarczynski (approved):
My PR LGTM XD

PR #114, opened 2023-06-09
Redesign new compositor callbacks API

WojciechBarczynski:
Suggestions that I came up with reviewing #113 and writing #112.


lib/membrane/video_compositor/handler/inputs_description.ex
```diff
@@ -0,0 +1,13 @@
...
+  Definition of all VC input videos used in composition.
+  """
+
+  alias Membrane.Pad
+  alias Membrane.VideoCompositor.Handler.InputProperties
+
+  @typedoc """
+  Describe all VC input videos used in composition.
+  """
+  @type t :: %{Pad.ref_t() => InputProperties.t()}
```
DominikWolek, line 12: NIT: this module is so small that `InputProperties` can be part of it to make it easier for reading.

lib/membrane/video_compositor/handler.ex
```diff
@@ -39,35 +39,29 @@ defmodule Membrane.VideoCompositor.Handler do
...
   Callback invoked upon initialization of Video Compositor.
   """
-  @callback handle_init() :: immediate_callback_return()
+  @callback handle_init(ctx :: CallbackContext.Init.t()) :: immediate_callback_return()
 
   @doc """
-  Callback invoked upon addition of the new
-  `Membrane.VideoCompositor.Scene.Object.InputVideo`
+  Callback invoked upon change of VC input videos.
+  Events changing input videos:
```
DominikWolek, line 46:
```suggestion
  Callback invoked upon change of VC input videos.
  
  Events changing input videos:
```

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -22,11 +22,37 @@ defmodule Membrane.VideoCompositor do
   @type update_transformations ::
           {:update_transformations, [{Membrane.Pad.ref_t(), VideoTransformations.t()}]}
 
+  @init_metadata_doc """
+  User-specified init metadata passed to handler callbacks.
+  Passing init metadata into `c:Membrane.VideoCompositor.Handler.handle_init/1` callback allows
+  the user to alternate custom-implemented init callback logic.
+  """
+
+  @typedoc @init_metadata_doc
```
DominikWolek, line 31: Why?

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -22,11 +22,37 @@ defmodule Membrane.VideoCompositor do
...
 
+  @input_pad_metadata_doc """
+  User-specified input stream metadata passed to handler callbacks.
+  Passing pad metadata into `c:Membrane.VideoCompositor.Handler.handle_inputs_change/3`
+  callback, allows the user to alternate custom-implemented callbacks logic,
+  e.g. prioritizing input stream in the `#{inspect(Scene)}` structs returned from callback.
+  """
+
+  @typedoc @input_pad_metadata_doc
+  @type input_pad_metadata :: any()
```
DominikWolek, line 54: As above

DominikWolek (changes requested):

lib/membrane/video_compositor/video_compositor.ex
WojciechBarczynski: Well, we use this attribute in two places. The first one is the `typedoc` and the second one is the `metadata option` `description`. 
I did it that way because: 
1. Having this description directly available in both places IMO makes it easier to read documentation than referencing the `typedoc` in the description.
2. Extracting `typedoc` in `description` using `#{inspect(Module.get_attribute/2)}` was painful. For me, it seemed to be an antipattern and over-engineering. 
3. Repeating this description twice obviously makes it harder to maintain and breaks DRY

lib/membrane/video_compositor/video_compositor.ex
WojciechBarczynski: As above

lib/membrane/video_compositor/handler/callback_context.ex
```diff
@@ -0,0 +1,23 @@
...
+  @enforce_keys [:scene, :inputs_description, :next_buffer_pts]
+  defstruct @enforce_keys
+
+  @typedoc """
+  Contains state of VC before event invoking of callback.
+  """
+  @type t :: %__MODULE__{
+          scene: Scene.t() | TemporalScene.t(),
+          inputs_description: InputsDescription.t(),
+          next_buffer_pts: Membrane.Time.non_neg_t()
```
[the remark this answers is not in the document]
WojciechBarczynski: I agree, `next_frame_pts` makes way more sense

lib/membrane/video_compositor/temporal_scene.ex
```diff
@@ -1,4 +1,4 @@
-defmodule Membrane.VideoCompositor.Scene.Temporal do
+defmodule Membrane.VideoCompositor.TemporalScene do
```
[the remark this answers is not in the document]
WojciechBarczynski: I don't like the proposed structure mainly, because when user alias it (with such long names, they probably will), they become unreadable.

Aliasing `Membrane.VideoCompositor.Scene.Temporal` means, that now all user `typespecs` use this weird `Tempotal.t()` type, that is hardly resembling `scene` hiding underneath it.

Using `Membrane.VideoCompositor.Temporal.Scene` means, that to avoid ambiguity, we have to `alias` with `as` one of: 
- `scene`
- `temporal scene`

(BTW while writing such code, I would probably alias it as `TemporalScene` :) )

Since `temporal scene` isn't a struct, we can stick with `Membrane.VideoCompositor.Scene.Temporal` if you want to (it's only a hindrance in `typespecs`, with the possibility of solving that using `Scene.Temporal.t()` in them).

Aside from that, I think that `temporal scene` is rather an alternative to `scene`, than its extension. Therefore, putting this module into `Membrane.VideoCompositor.TemporalScene` doesn't sound wrong to me.
 
Let me know what you think about it. It's up to you to decide whether we want `Membrane.VideoCompositor.TemporalScene` or `Membrane.VideoCompositor.Scene.Temporal`, I don't have a strong opinion on that. I just wanted to provide arguments for `Membrane.VideoCompositor.TemporalScene`, but if you still think `Membrane.VideoCompositor.Scene.Temporal` is better, we can use that.

lib/membrane/video_compositor/handler/inputs_description.ex
WojciechBarczynski: Good one

WojciechBarczynski:
After thoughtful consideration, I think we should enforce returning `{scene, state}` (remove `idle_callback_return`. Given typing `@type state :: any()` one may use `state = {scene, sth}` and we don't have any way to distinguish returning `state` from `{scene, state}` in such case. Moreover, it doesn't simplify anything so far, so I would enforce `{scene, state}` just like the membrane core does with `{actions, state}`

lib/membrane/video_compositor/handler/inputs.ex
```diff
@@ -0,0 +1,28 @@
...
+    alias Membrane.{RawVideo, VideoCompositor}
+
+    @enforce_keys [:stream_format, :metadata]
+    defstruct @enforce_keys
+
+    @type t :: %__MODULE__{
+            stream_format: RawVideo.t(),
+            metadata: VideoCompositor.input_pad_metadata()
+          }
+  end
```
DominikWolek, line 27: Why not `Input.Properties` structure?

lib/membrane/video_compositor/handler.ex
```diff
@@ -20,72 +20,55 @@ defmodule Membrane.VideoCompositor.Handler do
...
-              video :: video_details(),
-              ctx :: CallbackContext.VideoAdd.t(),
-              state :: state()
-            ) :: immediate_callback_return() | idle_callback_return()
+  Callback invoked upon change of VC input videos.
 
-  @doc """
-  Callback invoked upon removal of
-  `Membrane.VideoCompositor.Scene.Object.InputVideo`
+  Events changing input videos:
```
DominikWolek, line 43:
```suggestion
  `inputs` changing events:
```

DominikWolek: Or something like this, IMO it's more readable

lib/membrane/video_compositor/handler/inputs.ex
WojciechBarczynski: Moved `Inputs` definition to `Handler` and left `InputProperties`

lib/membrane/video_compositor/temporal_scene.ex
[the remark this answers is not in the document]
DominikWolek: After some discussion and consulting with @bblaszkow06 , this is what was suggested and what I think:

- There are no meaningful reasons for `VideoCompositor.TemporalScene` to be a separate module
-  Types in `VideoCompositor.TemporalScene` can be moved to `VideoCompositor.Handler` - with `VideoCompositor.TemporalScene.t()` moved to `VideoCompositor.Handler.temporal_scene()`
- `##Expiration` part of moduledoc should be moved to typedoc of `temporal_scene()`
- `##Examples` part of moduledoc should be moved to moduledoc of `Handler`

lib/membrane/video_compositor/handler.ex
```diff
@@ -2,90 +2,217 @@ defmodule Membrane.VideoCompositor.Handler do
...
 
-  Implementing handler allows to provide custom implementation and
+  Implementing a handler allows to provide custom implementation and
   react to various events, among others by setting a new scene and/or
   the inner custom state.
+
+  ## Examples
+
+  ### Simple video conference room
+  Video Compositor is used in video conference app like the videoroom.
```
DominikWolek, line 12:
```suggestion
  ### Video conference room
  Video Compositor is used in video conference app like the [Membrane Videoroom](https://github.com/membraneframework/membrane_videoroom).
```

lib/membrane/video_compositor/handler.ex
```diff
@@ -2,90 +2,217 @@ defmodule Membrane.VideoCompositor.Handler do
...
+
+    @impl true
+    def handle_info(msg, _ctx, _state) do
+      raise "Unsupported message \#{msg}"
+    end
+
+    @impl true
+    def handle_scene_expire(_expired_scene, _ctx, _state) do
+      raise "This shouldn't happen"
+    end
```
DominikWolek, line 42: - You can delete it from this example, IMO it only confuses the user
- We should create a separate task to add default callback implementation to avoid this

lib/membrane/video_compositor/handler.ex
```diff
@@ -2,90 +2,217 @@ defmodule Membrane.VideoCompositor.Handler do
...
+
+    @spec get_users_scene_mapping() :: %{(users_count :: non_neg_integer()) => Scene.t()}
+    defp get_users_scene_mapping() do
+      :do_stuff
+    end
+
+    @spec default_scene() :: Scene.t()
+    defp default_scene() do
+      :do_stuff
```
DominikWolek, line 51: Maybe something like this
```suggestion
      %Scene{
        objects: default_objects
        output: default_output
      }
```

lib/membrane/video_compositor/handler.ex
```diff
@@ -2,90 +2,217 @@ defmodule Membrane.VideoCompositor.Handler do
...
+    end
+
+    @impl true
+    def handle_scene_expire(_expired_scene, _ctx, _state) do
+      raise "This shouldn't happen"
+    end
+
+    @spec get_users_scene_mapping() :: %{(users_count :: non_neg_integer()) => Scene.t()}
+    defp get_users_scene_mapping() do
+      :do_stuff
```
DominikWolek, line 46: As above

lib/membrane/video_compositor/handler.ex
```diff
@@ -2,90 +2,217 @@ defmodule Membrane.VideoCompositor.Handler do
...
+
+  ## Examples
+
+  ### Simple video conference room
+  Video Compositor is used in video conference app like the videoroom.
+  Composition is determined by number of users.
+  This is how handler module might look like:
+
+  ```
+  defmodule ExampleHandler do
```
DominikWolek, line 17:
```suggestion
  defmodule ConferenceHandler do
```

lib/membrane/video_compositor/handler.ex
```diff
@@ -2,90 +2,217 @@ defmodule Membrane.VideoCompositor.Handler do
...
+  alias Membrane.{Pad, VideoCompositor}
   alias Membrane.VideoCompositor.Scene
 
   @typedoc """
   Type of user-managed inner state of the handler.
   """
   @type state :: any()
 
-  @type video_details :: {pad :: Pad.ref_t(), format :: StreamFormat.t()}
+  @typedoc """
```
DominikWolek, line 67: Should we mention that this is not implemented? What do you think?

DominikWolek (approved):

lib/membrane/video_compositor/handler.ex
WojciechBarczynski: This is merged to `new-compositor`. Nothing there works anyway XD

DominikWolek (approved):

PR #115, opened 2023-06-13
Ms 539 deserialization

jerzywilczek:
- change the transformations and layouts to use `PluginRegistryKey`s instead of `name`s, which were raw strings
- add a `PluginRegistry` struct
- keep both transformations and layouts in a single registry, since they're both identified by a registry key
- extract the logic common to layouts and transformations into a new trait, `CustomTransformer`
- attach a string to each `CustomStructElixirPacket`, which allows to check which transformer it is related to

### DIFF
--- a/native/membrane_video_compositor/src/compositor.rs
+++ b/native/membrane_video_compositor/src/compositor.rs
@@ -7,7 +7,7 @@ use membrane_video_compositor_common::plugins::layout::UntypedLayout;
 use membrane_video_compositor_common::plugins::transformation::{
     Transformation, UntypedTransformation,
 };
-use membrane_video_compositor_common::plugins::{CustomProcessor, PluginRegistryKey};
+use membrane_video_compositor_common::plugins::{PluginArgumentEncoder, PluginRegistryKey};
 use membrane_video_compositor_common::WgpuContext;
 
 use self::errors::CompositorError;
@@ -85,7 +85,7 @@ impl MockTransformation {
     const NAME: &str = "mock_transformation";
 }
 
-impl CustomProcessor for MockTransformation {
+impl PluginArgumentEncoder for MockTransformation {
     type Arg = String;
 
     fn registry_key() -> PluginRegistryKey<'static>
@@ -121,7 +121,7 @@ pub fn mock_transformation(ctx: StructElixirPacket<WgpuContext>) -> Transformati
 
 #[rustler::nif(schedule = "DirtyIo")]
 pub fn encode_mock_transformation(
-    arg: <MockTransformation as CustomProcessor>::Arg,
+    arg: <MockTransformation as PluginArgumentEncoder>::Arg,
 ) -> CustomStructElixirPacket {
     unsafe { MockTransformation::encode_arg(arg) }
 }
--- a/native/membrane_video_compositor/src/compositor.rs
+++ b/native/membrane_video_compositor/src/compositor.rs
@@ -64,35 +64,32 @@ impl InnerState {
         &mut self,
         transformation: Arc<dyn UntypedTransformation>,
     ) -> Result<(), CompositorError> {
-        if self
-            .plugin_registry
-            .contains_key(&transformation.registry_key())
-        {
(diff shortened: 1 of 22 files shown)


native/membrane_video_compositor/src/compositor.rs
```diff
@@ -60,14 +64,19 @@ impl InnerState {
         &mut self,
         transformation: Arc<dyn UntypedTransformation>,
     ) -> Result<(), CompositorError> {
-        if self.transformations.contains_key(transformation.name()) {
-            return Err(CompositorError::TransformationNameTaken(
-                transformation.name().to_string(),
+        if self
```
varsill, line 67: Perhaps we could gt rid of these repeatable code if we implemented some private function `register` and called it e.g. as: ` self.register(transformation.registry_key(), PluginRegistryEntry::Transformation(transformation))`

native/membrane_video_compositor/src/scene.rs
```diff
@@ -22,7 +24,8 @@ pub enum Node {
     Transformation {
         resolution: TextureOutputResolution,
         previous: Arc<Node>,
-        transformation: ImplementationPlaceholder,
+        // TODO: When we're ready, this should be swapped for `DecodedCustomStructElixirPacket`
+        transformation: CustomStructElixirPacket,
```
varsill, line 28: Perhaps we could think of a better name than `Transformation.transformation`.

jerzywilczek: I renamed it to `params`, same as in the layout

native/membrane_video_compositor_common/src/plugins.rs
```diff
@@ -1,2 +1,29 @@
+use crate::elixir_transfer::CustomStructElixirPacket;
+
 pub mod layout;
 pub mod transformation;
+
+#[derive(Debug, PartialEq, Eq, Hash)]
+pub struct PluginRegistryKey<'a>(pub &'a str);
+
+// Question for the reviewers: What should this trait be named? I don't like this name.
```
varsill, line 9: Quite a brave proposal from me: `Renderer` :D

native/membrane_video_compositor_common/src/plugins.rs
```diff
@@ -1,2 +1,29 @@
...
+pub trait CustomProcessor: Send + Sync + 'static {
+    type Arg: Send + 'static;
+
+    /// Should return the same thing as `name_dyn`
```
varsill, line 13: I guess there is no more `name_dyn` and `name`

native/membrane_video_compositor/src/compositor.rs
varsill, line 1: I am just wondering whether it wouldn't be better if encapsulated the whole logic concerning plugin registering in a separate struct `PluginRegistery`. Such a struct could contain a `HashMap` with some fixed types (`PluginRegistryKey` and `PluginRegistryEntry`). The `InnerState::register` function could be implemented for that struct. In case there would be a need of refering to the `HashMap` directly, we could implement some `Defer` trait, so that there wouldn't be a need to do `inner_state.plugin_registry.hash_map`, but I don't think it's necessary now.

jerzywilczek: I did it anyway

jerzywilczek: it clears up the logic nicely

native/membrane_video_compositor_common/src/plugins.rs
jerzywilczek: I don't really like that, because even though all layouts and transformations are necessarily renderers, this trait has nothing to do with rendering. none of it's methods have anything to do with rendering, they do however have a lot to do with it being custom and pluggable.

varsill (approved):
🥇

varsill (approved):

WojciechBarczynski (approved):

WojciechBarczynski:
Closing. We won't create plugin architecture right now.

PR #117, opened 2023-06-21
Design plugin's transformations API

noituri:


native/membrane_video_compositor/src/compositor.rs
```diff
@@ -97,8 +98,16 @@ impl PluginArgumentEncoder for MockTransformation {
...
 impl Transformation for MockTransformation {
-    fn do_stuff(&self, arg: &Self::Arg) {
-        println!("This is a mock transformation called with the string \"{arg}\" :^)")
+    type Error = ();
+
+    fn apply(
+        &self,
+        arg: &Self::Arg,
+        _source: &Texture,
+        _target: &Texture,
```
WojciechBarczynski, line 107: Shouldn't this one be mutable? I guess we want to "draw onto" this texture.
```suggestion
        _target: &mut Texture,
```

native/membrane_video_compositor/src/compositor.rs
WojciechBarczynski, line 1: [side note]
I don't think this is the best place for `MockTransformation`. 
This is temporary anyway, so there isn't a need to move it, but it would be nice if we keep responsibilities separated in the final version (like we didn't before) :).

WojciechBarczynski (approved):
Overall looks good to me.
Just make sure, what type of ref we need in those `apply` functions (I think that the mutable one would be necessary)

native/membrane_video_compositor/src/compositor.rs
WojciechBarczynski: Ok, it doesn't have to be

noituri: Draw methods do not require mutable reference

PR #119, opened 2023-06-30
Live EOS strategy

WojciechBarczynski:


Karolk99 (commented):
started_playing is only set when a pad is added what about the `start_composing` message? We can start playing without any pads.

lib/membrane/video_compositor/queue/strategy/live.ex
```diff
@@ -48,22 +49,31 @@ defmodule Membrane.VideoCompositor.Queue.Strategy.Live do
...
 
+  @impl true
+  def handle_playing(_ctx, state) do
+    {[stream_format: {:output, %CompositorCoreFormat{pad_formats: %{}}}], state}
+  end
+
   @impl true
   def handle_pad_added(pad, context, state) do
-    state = Bunch.Struct.put_in(state, [:pads_states, pad], PadState.new(context.options))
+    state =
```
Karolk99, line 73: I would put it in `handle_start_of_stream`

WojciechBarczynski: We need to have access to input pad options, to pass them into `handler`, which are only available in `handle_pad_added` and `handle_pad_removed`, so we need to "register" the input pad queue here or keep this information somehow. I changed the logic slightly, such that it should handle unlinking pads without sending any buffers.

Karolk99 (approved):

PR #129, opened 2023-09-19
Add request adapter logging

WojciechBarczynski:
Logging for CI failing adapter request

### DIFF
--- a/native/membrane_videocompositor/src/elixir_bridge.rs
+++ b/native/membrane_videocompositor/src/elixir_bridge.rs
@@ -197,7 +197,9 @@ fn set_videos(
 }
 
 fn load(env: Env, _: Term) -> bool {
-    env_logger::init();
+    env_logger::init_from_env(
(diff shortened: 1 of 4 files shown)


native/membrane_videocompositor/src/elixir_bridge.rs
```diff
@@ -196,11 +196,15 @@ fn set_videos(
     Ok(atoms::ok())
 }
 
+fn load(env: Env, _: Term) -> bool {
+    env_logger::init();
```
wkozyra95, line 200: It's producing a lot of logs, it might not be desired effect to enable it by default if someone is using it on prod.  You can add
```
    env_logger::init_from_env(
        env_logger::Env::default().filter_or(env_logger::DEFAULT_FILTER_ENV, "warn"),
    );
```
and set RUST_LOG=info on CI

wkozyra95 (approved):

PR #131, opened 2023-09-23
VideoCompositor Membrane SDK

WojciechBarczynski:
Membrane SDK for VideoCompositor


lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -0,0 +1,431 @@
...
+    state = %State{
+      state
+      | inputs: [
+          %InputState{input_id: input_id, port_number: input_port, pad_ref: input_ref} | inputs
+        ]
+    }
+
+    spec =
+      bin_input(Pad.ref(:input, pad_id))
+      |> via_in(Pad.ref(:input, pad_id),
```
FelonEkonom, line 182:
```suggestion
      bin_input(input_ref)
      |> via_in(input_ref,
```

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -0,0 +1,431 @@
...
+          | outputs
+        ]
+    }
+
+    spec =
+      child(Pad.ref(:upd_source, pad_id), %UDP.Source{
+        local_port_no: port,
+        local_address: @local_host,
+        recv_buffer_size: @udp_buffer_size
+      })
```
FelonEkonom, line 223: Using `Pad.ref` to name a child is weird, please change it 🙈 🙈 🙈 You can name this child just `{:udp_source, pad_id}`

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -0,0 +1,431 @@
...
+  @impl true
+  def handle_pad_removed(input_ref = Pad.ref(:input, pad_id), _ctx, state = %State{}) do
+    state = remove_input(state, input_ref)
+    {[remove_child: [{:rtp_sender, pad_id}, {:upd_sink, pad_id}]], state}
+  end
+
+  @impl true
+  def handle_pad_removed(output_ref = Pad.ref(:output, pad_id), _ctx, state = %State{}) do
+    state = remove_output(state, output_ref)
+    {[remove_child: [{:rtp_receiver, pad_id}, {:upd_source, pad_id}]], state}
```
FelonEkonom, line 240: Name of this child in line 219 was `Pad.ref(...)`, instead of tuple with `:udp_source`. Have you tested this code or checked anyhow if it works?

lib/membrane/video_compositor/request.ex
```diff
@@ -0,0 +1,162 @@
...
+    vc_url = ip_to_url(@local_host, vc_port)
+
+    req_result =
+      Req.post(vc_url,
+        json: %{
+          type: "start"
+        }
+      )
+
+    handle_req_result(req_result)
```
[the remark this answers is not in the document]
FelonEkonom: You can change it in the rest of the places in this file as well

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -0,0 +1,431 @@
...
+
+    spawn(fn ->
+      Rambo.run(vc_app_path, [], env: %{"MEMBRANE_VIDEO_COMPOSITOR_API_PORT" => "#{vc_port}"})
+    end)
+
+    started? =
+      0..50
+      |> Enum.reduce_while(false, fn _i, _acc ->
+        sleep_time_ms = 100
+        :timer.sleep(sleep_time_ms)
```
FelonEkonom, line 323: use `Process.sleep`

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -0,0 +1,431 @@
...
+
+        case VcReq.send_custom_request(%{}, vc_port) do
+          {:ok, _} -> {:halt, true}
+          {:error, _} -> {:cont, false}
+        end
```
FelonEkonom, line 328: Are you sure you don't want to check if `response.status` is 200?

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -0,0 +1,431 @@
...
+      end)
+
+    unless started? do
+      raise "Failed to startup and connect to VideoCompositor server."
+    end
+
+    :ok
+  end
```
FelonEkonom, line 336: `VcReq.send_custom_request` is the only function from `VcReq` module, that returns an `:ok` tuple if `response.status != 200`. You check the `response.status` in line 251, but not in `case` in line 325.  Are you sure, that this is ok? For me it looks little inconsistent and opens a place for easy mistakes, so it would be good to refactor it

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -0,0 +1,431 @@
...
+            raise "Unsupported system architecture: #{system_architecture}"
+        end
+
+      {:unix, :linux} ->
+        :linux_x86_64
+
+      os_type ->
+        raise "Unsupported os type: #{os_type}"
+    end
+  end
```
FelonEkonom, line 361: 1) This same issue as in `Mix.Tasks.DownloadCompositor`
2) You have equal private function in `Mix.Tasks.DownloadCompositor`. Don't repeat yourself and make this function public, maybe extract it to another module

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -0,0 +1,431 @@
...
+            input_id,
+            state,
+            input_port + 2
+          )
+
+        _other ->
+          raise "Register input failed"
+      end
+    end
+  end
```
FelonEkonom, line 384: Will this function be called many times? If yes, you will check port `4000` many times. This could be fixed by adding some randomisation.

BTW, why the first port, that you try to open, is 4000? It doesn't sound like a good idea

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -0,0 +1,431 @@
...
+
+    %State{state | inputs: inputs}
+  end
+
+  @spec remove_output(State.t(), Membrane.Pad.ref()) :: State.t()
+  defp remove_output(state = %State{outputs: outputs}, output_ref) do
+    output_id =
+      outputs
+      |> Enum.find(fn %OutputState{pad_ref: ref} -> ref == output_ref end)
+      |> then(fn %OutputState{output_id: id} -> id end)
```
FelonEkonom, line 423: Use `Map.get` instead. Do it here and in similar places

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -0,0 +1,431 @@
...
+      end)
+
+    unless started? do
+      raise "Failed to startup and connect to VideoCompositor server."
+    end
+
+    :ok
+  end
+
+  @spec system_architecture() :: :darwin_aarch64 | :darwin_x86_64 | :linux_x86_64
```
FelonEkonom, line 338:
```suggestion
  @spec system_architecture() :: :darwin_aarch64 | :darwin_x86_64 | :linux_x86_64 | no_return()
```

FelonEkonom (changes requested):

lib/membrane/video_compositor/download_compositor.ex
```diff
@@ -16,8 +16,12 @@ defmodule Mix.Tasks.DownloadCompositor do
 
     path = File.cwd!() |> Path.join("video_compositor_app/#{vc_architecture}")
 
-    File.mkdir_p!(path)
-    _wget_res = "wget -nc #{url} -O - | tar -xvz -C #{path}" |> String.to_charlist() |> :os.cmd()
+    unless File.exists?(path) do
+      File.mkdir_p!(path)
+
+      _wget_res =
+        "wget -nc #{url} -O - | tar -xvz -C #{path}" |> String.to_charlist() |> :os.cmd()
```
FelonEkonom, line 23: 1) Use `System.cmd` instead of `:os.cmd`
2) Don't ignore return value of this command

FelonEkonom (changes requested):

examples/layout_with_shader/pipeline.ex
```diff
@@ -0,0 +1,140 @@
...
+      |> child(:sdl_player, Membrane.SDL.Player)
+    ]
+
+    Process.send_after(self(), :register_shader, 10)
+    Process.send_after(self(), :add_input, 5000)
+    Process.send_after(self(), :add_input, 10_000)
+    Process.send_after(self(), :add_input, 15_000)
+    Process.send_after(self(), :add_input, 20_000)
+
+    {[spec: spec, spec: spec_2], %{videos_count: 1}}
```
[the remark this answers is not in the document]
WojciechBarczynski: `Membrane.Core` changes element mode/state to `playing` when the whole spec is processed . Without linking outputs, the spec isn't considered processes, so VC doesn't go into `playing` mode. Outputs need to receive RTP streams to link RTP.SessionBin to VideoCompositor bin output. In the case of spawning VideoCompositor in a single spec with inputs/outputs, VC won't go into a `playing` state linking outputs, that won't link before receiving RTP streams, that won't be sent by VideoCompsitor, since it's not in `playing` mode. Basically, it's a classical deadlock.
I'll try to document this better.

lib/membrane/video_compositor/video_compositor.ex
WojciechBarczynski: Yep, this is basically pinging the VideoCompositor server to check whether it's already up. I'll make this more readable.

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -0,0 +1,472 @@
...
+                spec: :on_init | :on_message,
+                description:
+                  "Specifies when VideoCompositor starts composing frames. In `:on_message` strategy, `:start_composing` message have to be send to start composing.",
+                default: :on_init
+              ],
+              vc_server_port_number: [
+                spec: non_neg_integer(),
+                description:
+                  "Port on which VC server should run. Port have to be unused. In case of running multiple VC elements, those values should be unique.",
+                default: 8001
```
mat-hek, line 157: choosing a random port by default would be more appropriate here

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -0,0 +1,472 @@
...
+  @impl true
+  def handle_init(_ctx, opt) do
+    vc_port = opt.vc_server_port_number
+    :ok = start_vc_server(vc_port)
+
+    :ok = VcReq.init(opt.framerate, opt.stream_fallback_timeout, opt.init_web_renderer?, vc_port)
+
+    if opt.start_composing_strategy == :on_init do
+      :ok = VcReq.start_composing(vc_port)
+    end
```
mat-hek, line 199: since this may take some time, let's do it in `handle_setup`

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -0,0 +1,472 @@
...
+                default: 8001
+              ]
+
+  def_input_pad :input,
+    accepted_format: %Membrane.H264{alignment: :nalu, stream_structure: :annexb},
+    availability: :on_request,
+    options: [
+      input_id: [
+        spec: input_id(),
+        description: "Input identifier."
```
mat-hek, line 166: yes

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -0,0 +1,472 @@
...
+      )
+    end
+
+    register_input_or_output(try_register, state)
+  end
+
+  @spec register_input_or_output((:inet.port_number() -> VcReq.req_result()), State.t()) ::
+          {:ok, :inet.port_number()} | :error
+  defp register_input_or_output(try_register, state) do
+    6000..8000
```
mat-hek, line 421: this should be configurable, or at least a module attribute

lib/membrane/video_compositor/state.ex
```diff
@@ -0,0 +1,34 @@
...
+
+  @spec used_ports(t()) :: MapSet.t()
+  def used_ports(%__MODULE__{inputs: inputs, outputs: outputs, vc_port: vc_port}) do
+    input_ports =
+      inputs |> Enum.map(fn input_state -> input_state.port_number end) |> MapSet.new()
+
+    output_ports =
+      outputs |> Enum.map(fn output_state -> output_state.port_number end) |> MapSet.new()
+
+    MapSet.new([vc_port]) |> MapSet.union(input_ports) |> MapSet.union(output_ports)
```
mat-hek, line 32: Can't it be just

```suggestion
      [vc_port] ++ Enum.map(inputs ++ outputs, & &1.port_number)
```

mat-hek (changes requested):

lib/membrane/video_compositor/download_compositor.ex
```diff
@@ -0,0 +1,62 @@
...
+    end
+  end
+
+  @spec vc_app_path() :: String.t()
+  def vc_app_path() do
+    File.cwd!() |> Path.join("#{vc_app_directory()}/video_compositor/video_compositor")
+  end
+
+  defp vc_app_directory() do
+    "video_compositor_app/#{@vc_version}/#{system_architecture()}"
```
mat-hek, line 35: This should probably be downloaded into `priv`, now it won't be available when deployed

mat-hek (changes requested):

lib/membrane/video_compositor/video_compositor.ex
WojciechBarczynski: I don't understand what you mean. 
- If you suggest using `Map.get()` to find `OutputState` with `pad_ref == ouptut_ref` I would have to first convert the list to map to just perform `get` on it (I don't want to keep this as map since I perform `find` on different fields, convert this to `context` etc.).  I think that would be less readable.
- If you suggest changing the `then` section to `Map.get()` I'm also against it. I would remove the compile time check of the `output_id` field name, making the code more error-prone.

Regardless of which line you suggested to change, I don't think it's a good idea.

WojciechBarczynski: Matching on the other side is better. I will use that.

mix.exs
```diff
@@ -31,12 +32,35 @@ defmodule Membrane.Template.Mixfile do
...
+      {:membrane_udp_plugin, "~> 0.10.0"},
+      # VC server start
+      {:rambo, "~> 0.3"},
+      # VC API requests
+      {:req, "~> 0.4.0"},
+      {:jason, "~> 1.4"},
+      # Examples
+      {:membrane_file_plugin, "~> 0.15.0"},
+      {:membrane_realtimer_plugin, "~> 0.7.0"},
+      {:membrane_sdl_plugin, "~> 0.16.0"},
```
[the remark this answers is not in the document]
WojciechBarczynski: I tried it in multiple configurations, but I couldn't make it work. 
I got multiple different errors, but the most frequent one was:`Unchecked dependencies for environment prod:`

lib/membrane/video_compositor/compile/download_compositor.ex
```diff
@@ -0,0 +1,66 @@
...
+  @vc_version "v0.1.0-rc.2"
+  @membrane_video_compositor_plugin_path File.cwd!()
+
+  @impl Mix.Task
+  def run(_args) do
+    url =
+      "https://github.com/membraneframework/video_compositor/releases/download/#{@vc_version}/video_compositor_#{system_architecture()}.tar.gz"
+
+    unless File.exists?(vc_app_directory()) do
+      File.mkdir_p!(vc_app_directory())
```
wkozyra95, line 17:
```suggestion
    unless File.exists?(Path.join(vc_app_directory(), ".lock")) do
      File.rm_rf!(vc_app_directory())
      File.mkdir_p!(vc_app_directory())
```
and create that `.lock` file at the end

lib/membrane/video_compositor/output_options.ex
```diff
@@ -0,0 +1,35 @@
...
+          | :faster
+          | :fast
+          | :medium
+          | :slow
+          | :slower
+          | :veryslow
+          | :placebo
+
+  @enforce_keys [:width, :height, :id]
+  defstruct @enforce_keys ++ [encoder_preset: :fast]
```
wkozyra95, line 24: is the fast a default? I would prefer not to introduce any default values in a client SDK, if we want to have a default it should be in a compositor directlly

lib/membrane/video_compositor/request.ex
```diff
@@ -0,0 +1,122 @@
...
+  @type request_result ::
+          {:ok, Req.Response.t()}
+          | {:error_response_code, Req.Response.t()}
+          | {:error, Exception.t()}
+
+  @spec init(non_neg_integer(), Membrane.Time.t(), boolean(), :inet.port_number()) ::
+          request_result()
+  def init(framerate, stream_fallback_timeout, init_web_renderer?, vc_port) do
+    %{
+      type: "init",
```
wkozyra95, line 19: not sure if it's idiomatic in elixir, but wouldn't field like type be a good candidate to use atoms instead of strings?

examples/lib/dynamic_outputs/pipeline.ex
```diff
@@ -0,0 +1,194 @@
...
+
+    [notify_child: {:video_compositor, {:vc_request, update_scene_request}}]
+  end
+
+  @spec tiled_layout(list(String.t())) :: map()
+  defp tiled_layout(input_pads) do
+    %{
+      type: "built-in",
+      transformation: :tiled_layout,
+      node_id: "layout",
```
WojciechBarczynski, line 151: I wouldn't say I like mixing strings and atoms like this. When we release VC in the new version, where `"builtin"` doesn't use `"-"` in the middle, I'll replace all `type:` fields with atoms. Leaving `node_id` as a string is a good idea since they shouldn't be dynamically created in runtime (they aren't garbage collected, the beam has a limit on the number of created atoms, etc.).

README.md
```diff
@@ -1,27 +1,45 @@
...
 
-The package can be installed by adding `membrane_template_plugin` to your list of dependencies in `mix.exs`:
+The package can be installed by adding `membrane_video_compositor_plugin` to your list of dependencies in `mix.exs`:
 
 ```elixir
 def deps do
   [
-    {:membrane_template_plugin, "~> 0.1.0"}
+    {:membrane_video_compositor_plugin, "~> 0.5.4"},
+    {:membrane_rtp_h264_plugin, "~> 0.18.0"}
```
WojciechBarczynski, line 19: Without `:membrane_rtp_h264_plugin` `RTP.SessionBin` can't resolve the payload type of `encoding: :H264`. IDK why it doesn't work without listing it in deps, since VC plugin already has a dependency to this package, but I'll try to figure it out later on.

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -231,242 +344,194 @@ defmodule Membrane.VideoCompositor do
...
+      state
+      | outputs: [output_state | outputs]
     }
 
     spec =
-      get_child({:rtp_receiver, pad_id})
-      |> via_out(Pad.ref(:output, ssrc), options: [depayloader: RTP.H264.Depayloader])
-      |> child({:output_processor, pad_id}, %Membrane.VideoCompositor.OutputProcessor{
-        output_stream_format: output_stream_format
+      child({:upd_source, id}, %UDP.Source{
```
mat-hek, line 465:
```suggestion
      child({:udp_source, id}, %UDP.Source{
```

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -231,242 +344,194 @@ defmodule Membrane.VideoCompositor do
...
 
     output_children = [
-      {:rtp_receiver, pad_id},
-      {:upd_source, pad_id},
-      {:rtp_receiver, pad_id},
-      {:output_processor, pad_id}
+      {:rtp_receiver, output_id},
+      {:upd_source, output_id},
+      {:rtp_receiver, output_id},
+      {:output_processor, output_id}
```
mat-hek, line 432: instead of listing the children, use children groups

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -102,120 +111,224 @@ defmodule Membrane.VideoCompositor do
...
+            |> Enum.take_random(1)
 
-    :ok = VcReq.init(opt.framerate, opt.stream_fallback_timeout, opt.init_web_renderer?, vc_port)
+          port
+
+        port when is_integer(port) ->
+          port
+      end
+
+    :ok = ServerRunner.start_vc_server(vc_port)
```
mat-hek, line 301: We won't retry if the randomly chosen port is busy. Not sure how problematic this is.

mat-hek (commented):
Looks good, left minor requests

jerzywilczek (approved):
Didn't read the examples, the rest looks good

examples/lib/dynamic_outputs/pipeline.ex
```diff
@@ -0,0 +1,194 @@
...
+    # VideoCompositor forbids removing output used in the current scene.
+    {update_scene_action(inputs, outputs) ++ [remove_children: children],
+     %{state | outputs: outputs}}
+  end
+
+  @spec update_scene_action(list(VideoCompositor.input_id()), list(VideoCompositor.output_id())) ::
+          [Membrane.Pipeline.Action.notify_child()]
+  defp update_scene_action(input_ids, output_ids) do
+    update_scene_request =
+      if not Enum.empty?(output_ids) and not Enum.empty?(input_ids) > 0 do
```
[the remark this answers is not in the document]
WojciechBarczynski: If any of the specified nodes are unused for composition, then VideoCompositor returns the `400` code in the request-response. VideoCompositor defines node unused, one it doesn't provide any texture used in the output composition (basically runs DFS from output nodes and checks if all nodes have been reached). When there are no outputs, then this code would still define a single node that can't be used, therefore VC would return an error. It can be avoided with a separate function, but it just moves the if statement to a different place.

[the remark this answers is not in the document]
mat-hek: If I'm not mistaken, my suggestion would not change the behaviour of the code. But it's a minor thing, you can leave it as is

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -0,0 +1,537 @@
...
+
+  def_options framerate: [
+                spec: Membrane.RawVideo.framerate_t(),
+                description: "Framerate of VideoCompositor outputs."
+              ],
+              port_range: [
+                spec: port_range(),
+                description: """
+                Port range in which input and output streams would try to be registered.
+                If all ports in range will be used, VideoCompositor will crash on input/output registration.
```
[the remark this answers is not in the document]
WojciechBarczynski: It just thorws match error on register, su it's recoverable in Elixir.

WojciechBarczynski:
Since this is already quite big, I'll bump the membrane core version to `1.0` and the VideoCompositor version in follow-up PR, before releasing it.

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -534,4 +533,25 @@ defmodule Membrane.VideoCompositor do
 
     {[], state}
   end
+
+  @spec try_starting_on_port(:inet.port_number(), String.t()) ::
+          {:halt, {:ok, :inet.port_number()}} | {:cont, err :: String.t()}
+  defp try_starting_on_port(port, err) do
+    Membrane.Logger.info("Trying to lunch VideoCompositor on port: #{port}")
```
mat-hek, line 540: omnomnom

mat-hek: this should be debug BTW

examples/lib/dynamic_outputs/pipeline.ex
```diff
@@ -0,0 +1,195 @@
...
+    end
+
+    {[], state}
+  end
+
+  @impl true
+  def handle_child_notification(notification, child, _ctx, state) do
+    Membrane.Logger.info(
+      "Received notification: #{inspect(notification)} from child: #{inspect(child)}."
+    )
```
mat-hek, line 85: this should rather be a debug

examples/lib/layout_with_shader/pipeline.ex
```diff
@@ -0,0 +1,211 @@
...
+      """
+    end
+
+    {[], state}
+  end
+
+  @impl true
+  def handle_child_notification(notification, child, _ctx, state) do
+    Membrane.Logger.info(
+      "Received notification: #{inspect(notification)} from child: #{inspect(child)}."
```
mat-hek, line 101: debug again

mat-hek (approved):

WojciechBarczynski:
I'll address any additional comments in the follow up PR

PR #141, opened 2023-11-16
Release membrane_video_compositor_plugin v0.6.0

FelonEkonom:
### Release "Partly Solid Ghoul"

| Project | New Version |
|---|---|
| membrane_core | `1.0.0` |

<sup>proudly automated with <a href="https://github.com/membraneframework/sebex">Sebex</a></sup>

PR #142, opened 2023-11-22
update to core 1.0

mat-hek:


mat-hek:
@WojciechBarczynski it seems that elixir_test_with_software_rendering uses a docker image with stale elixir, thus dependencies don't compile. Can you update the elixir version there?

WojciechBarczynski (approved):

WojciechBarczynski:
1. Should #141 be closed?
2. I can merge and release it with failing CI and fix this in follow-up, so you don't have to wait to bump other packages.

PR #145, opened 2023-12-04
Wait with input registered notification for first input stream frame

WojciechBarczynski:


lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -175,7 +175,8 @@ defmodule Membrane.VideoCompositor do
           {:vc_request_response, request_body(), Req.Response.t(), Context.t()}
 
   @typedoc """
-  Notification send to parent after input registration.
+  Notification send to parent after VideoCompositor receives
```
wkozyra95, line 178:
```suggestion
  Notification sent to parent after VideoCompositor receives
```
the same issue in other comments

wkozyra95 (approved):

PR #152, opened 2023-12-11
Running video compositor from path

WojciechBarczynski:
Question for reviewers:
Idk where to put documentation about it. It should be purely for development purposes. Maybe some additional file like `DEVELOPMENT.md`? Any suggestions?


WojciechBarczynski:
Different approach for this in next PR

PR #158, opened 2024-01-25
update to latest compositor

wkozyra95:


lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -293,6 +288,17 @@ defmodule Membrane.VideoCompositor do
 
   @impl true
   def handle_setup(_ctx, opt) do
+    env = %{
```
wkozyra95, line 291: I don't like the fact that some options are only applied if compositor is starting the server. They will be silentlly ignored when compositor was already started. 

Note that this is not a new issue, current implementation has the same problem.

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -1,19 +1,18 @@
 defmodule Membrane.VideoCompositor do
   @moduledoc """
-  Membrane SDK for [VideoCompositor](https://github.com/membraneframework/video_compositor),
```
wkozyra95, line 3: Fixed some stuff that I noticed, but those docs could use another iteration, to both fix the language, but also to improve the explanations themselves.

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -240,7 +233,7 @@ defmodule Membrane.VideoCompositor do
                 description: """
                 Timeout that defines when the VideoCompositor should switch to fallback on the input stream that stopped sending frames.
                 """,
-                default: Membrane.Time.seconds(10)
+                default: Membrane.Time.seconds(2)
```
wkozyra95, line 236: the same value as in compositor

WojciechBarczynski (approved):

PR #160, opened 2024-02-01
Require explicit output ports

wkozyra95:
- remove `port_range` option
- add `api_port` option that can take ranger or specific port
- add `port` option to the input pad that can take range or a specific port
- add `port` field to register_request
- fix `env` object to use strings as keys instead of atoms
- rename `lc_server_config` to `server_setup`
- remove the option to enable Chrome support
  - I don't want to force everyone to use binary with Chrome, so I prefer to disable that option for now.


lib/membrane/video_compositor/server_runner.ex
```diff
@@ -1,10 +1,41 @@
...
+  require Membrane.Logger
+
+  alias Membrane.LiveCompositor
   alias Membrane.LiveCompositor.Request
 
-  @spec start_lc_server(:inet.port_number(), map()) :: :ok | :error
-  def start_lc_server(lc_port, env) do
+  @spec start_server(:inet.port_number() | LiveCompositor.port_range(), map()) ::
+          {:ok, :inet.port_number()} | {:error, err :: String.t()}
+  def start_server(port_or_port_range, env) do
```
wkozyra95, line 11: video_compositor.ex is very large now, and this code was easy to extract

lib/membrane/video_compositor/output_options.ex
```diff
@@ -30,6 +30,7 @@ defmodule Membrane.LiveCompositor.OutputOptions do
           width: Membrane.RawVideo.width_t(),
           height: Membrane.RawVideo.height_t(),
           id: Membrane.LiveCompositor.output_id(),
+          port: :inet.port_number(),
```
wkozyra95, line 33: if we simplify the pipeline setup this option will go to the pad

lib/membrane/video_compositor/request.ex
```diff
@@ -22,16 +22,20 @@ defmodule Membrane.LiveCompositor.Request do
...
       type: :register,
-      entity_type: :input_stream,
+      entity_type: :rtp_input_stream,
       input_id: "#{input_id}",
-      port: input_port_number,
+      port:
+        case port do
+          {start, endd} -> "#{start}:#{endd}"
+          exact_port -> exact_port
+        end,
```
WojciechBarczynski, line 38: I think this looks better than putting the case statement inside the map declaration
```elixir
port = case port do ...
```

```elixir
%{
  port: port
}
```

WojciechBarczynski (approved):
LGTM

PR #168, opened 2024-03-19
Use instance_id to identify if connecting to the correct instance

wkozyra95:


examples/lib/dynamic_outputs.exs
```diff
@@ -164,7 +164,9 @@ defmodule DynamicOutputsPipeline do
       },
       children:
         ctx.video_inputs
-          |> Enum.map(fn input_id -> %{type: :input_stream, input_id: input_id, id: "#{output_id}_#{input_id}"} end)
+        |> Enum.map(fn input_id ->
+          %{type: :input_stream, input_id: input_id, id: "#{output_id}_#{input_id}"}
+        end)
```
WojciechBarczynski, line 169: If outputs are named like `output_{n}` (as we do here), it means that components representing input k for output n will have id: "output_n_input_k".

I wouldn't say I like prefixing component names with output_id. Error messages should include information to which output they refer. I guess that you used it because the compositor requires component ids uniqueness across all outputs, but I think that shouldn't be the case. Dropping this check in favor of checking only single outputs' component IDs' uniqueness improves user experience in my opinion. This can however create issues in conversions between output component trees in the API layer to DAGs nodes in the renderer layer. I think we can use this to allow reuseing components across outputs (only if all of their children are the same, in other cases, we should somehow differentiate between them and render them separately).

I know that this is probably outside of the scope of this PR and I'm not sure if this has high priority, so you can just some different naming here and create an issue in the LC repo. Maybe sth like `{input_id}_for_{output_id}`

lib/membrane/video_compositor/server_runner.ex
```diff
@@ -101,15 +108,19 @@ defmodule Membrane.LiveCompositor.ServerRunner do
...
 
       with {:is_alive, true} <- {:is_alive, Process.alive?(pid)},
-           {:ok, _} <- Request.get_status(lc_port) do
-        {:halt, :started}
+           {:ok, response} <- Request.get_status(lc_port) do
+        if response.body["instance_id"] == instance_id do
+          {:halt, :started}
+        else
+          {:halt, :not_started}
+        end
```
WojciechBarczynski, line 123: Either log the error reason here or return the different value and log it in this function caller. The default error message wouldn't be helpful.

PR #169, opened 2024-03-25
Add websocket conection

wkozyra95:
- Connect to websocket in handle_setup
- `handle_setup` returns `:incomplete` to avoid into switching to playing state.
- when websocket is connected change to complete. This avoids race condition between conecting inputs streams and connecting to a websocket
- move start composing to after connection is established

### DIFF
--- a/lib/membrane/video_compositor/video_compositor.ex
+++ b/lib/membrane/video_compositor/video_compositor.ex
@@ -84,10 +84,9 @@ defmodule Membrane.LiveCompositor do
 
   alias Membrane.{Opus, Pad, RemoteStream, RTP, TCP}
 
-  alias Membrane.LiveCompositor
-
   alias Membrane.LiveCompositor.{
     Context,
+    EventHandler,
     Request,
     ServerRunner,
     State,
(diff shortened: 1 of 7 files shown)


lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -84,6 +84,8 @@ defmodule Membrane.LiveCompositor do
 
   alias Membrane.{Opus, Pad, RemoteStream, RTP, TCP}
 
+  alias Membrane.LiveCompositor
```
WojciechBarczynski, line 87: Aliasing the same module seems weird. You can either alias `EventHandler` directly or use `__MODULE__`

lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -742,6 +746,20 @@ defmodule Membrane.LiveCompositor do
...
+      {:ok, _resp} = Request.start_composing(state.lc_port)
+    end
+
+    {[setup: :complete], state}
+  end
+
+  @impl true
+  def handle_info({:websocket_message, {event_type, event_data}}, _ctx, state) do
+    {[notify_parent: {event_type, event_data, state.context}], state}
+  end
```
WojciechBarczynski, line 761: No docs for this behaviour

WojciechBarczynski (approved):

PR #170, opened 2024-03-27
Change ahead of time processing option to offline_processing

wkozyra95:


lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -236,17 +236,21 @@ defmodule Membrane.LiveCompositor do
...
-                stream before starting; otherwise, the compositor will run in a busy loop processing
-                data far into the future.
+                - `:offline_processing`
+                  - Output streams will be produced faster than in real time if input streams are
+                  ready.
+                  - Never drop output frames, even if the encoder or rendering process is not able to
+                  process data in real time.
+
+                    When using this option, make sure to register the output stream before starting;
+                    otherwise, the compositor will run in a busy loop processing data far into the future.
```
WojciechBarczynski, line 253: Tbh sounds like we shouldn't do that :)

WojciechBarczynski (approved):

lib/membrane/video_compositor/video_compositor.ex
wkozyra95: we don't have a good way to stop that without disabling valid use cases. We can discuss that offline in more detail if you want, but for now I'm merging this.

PR #172, opened 2024-03-27
Add option to pass raw FFmpeg encoder options

wkozyra95:


lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -383,6 +383,13 @@ defmodule Membrane.LiveCompositor do
         spec: video_encoder_preset(),
         default: :fast
       ],
+      ffmpeg_options: [
+        spec: map() | nil,
```
WojciechBarczynski, line 387: https://hexdocs.pm/elixir/typespecs.html#literals
```suggestion
        spec: %{String.t() => String.t()} | nil,
```
Technically atoms will work also since they are serialized to JSON strings via `Req` (or more specifically `jason` under the hood), but doing typespec for that would require defining a separate type.

WojciechBarczynski (approved):

PR #174, opened 2024-03-28
Update LiveCompositor binary add Linux ARM support

wkozyra95:


lib/membrane/video_compositor/compile/download_compositor.ex
```diff
@@ -43,26 +43,46 @@ defmodule Mix.Tasks.Compile.DownloadCompositor do
...
-        "linux_x86_64"
+      {:darwin, :x86} ->
+        "darwin_x86_64"
+
+      {:darwin, :aarch64} ->
+        "darwin_aarch64"
+
+      {system, arch} ->
+        Membrane.Logger.warning("Unsupported platform: #{system} #{arch}")
+    end
```
WojciechBarczynski, line 70: This doesn't match the function spec. Either change the spec and handle it there or raise it here (then you need to also add `no_return` to spec)

WojciechBarczynski (approved):

PR #176, opened 2024-04-15
Update to new API

wkozyra95:


examples/lib/layout_with_shader.exs
```diff
@@ -134,6 +135,15 @@ defmodule LayoutWithShaderPipeline do
...
   end
 
+  @impl true
+  def handle_child_notification(notification, child, _ctx, state) do
+    Membrane.Logger.debug(
+      "Received notification: #{inspect(notification)} from child: #{inspect(child)}."
+    )
+
+    {[], state}
+  end
```
WojciechBarczynski, line 145: How this even works? Those two `handle_child_notification` below won't match. Definition order matters in elixir due to pattern match and this matches all cases

PR #178, opened 2024-04-24
Enforce `{:ok, _}` result from init_requests

wkozyra95:


lib/membrane/video_compositor/video_compositor.ex
```diff
@@ -427,8 +427,9 @@ defmodule Membrane.LiveCompositor do
 
     opt.init_requests
     |> Enum.each(fn request ->
-      IntoRequest.into_request(request)
-      |> ApiClient.send_request(lc_port)
+      {:ok, _} =
+        IntoRequest.into_request(request)
+        |> ApiClient.send_request(lc_port)
```
WojciechBarczynski, line 432: [NIT]
I prefer to pattern match on this and raise more descriptive errors than " match errors " in such cases.

If you don't like formatting, you can pipe this result into:
```elixir
|> then(fn res -> 
           case res do
                 {:ok, _} -> ...
                 {:error, ...} -> raise "..."
           end 
      end
)
```

PR #186, opened 2024-04-30
Update README

wkozyra95:


WojciechBarczynski (commented):
It's not pretty, but as a temporary solution it's ok

PR #189, opened 2024-05-13
Send BYE packet on EOS

wkozyra95:
If we send BYE packet on EOS and avoid closing the socket, then we don't have to worry about timeouts on closed socket.


lib/membrane/live_compositor/rtcp_bye_sender.ex
```diff
@@ -0,0 +1,47 @@
...
+  def_options ssrc: [
+                spec: non_neg_integer()
+              ]
+
+  def_input_pad :input,
+    accepted_format: %RemoteStream{type: :packetized, content_format: RTP},
+    flow_control: :auto
+
+  def_output_pad :output,
+    accepted_format: %RemoteStream{type: :packetized, content_format: RTP},
```
wkozyra95, line 17: I'm not mentioning RTCP here intentionally

WojciechBarczynski (approved):

PR #194, opened 2024-06-11
Link to LiveCompositor license

wkozyra95:


darthez (approved):

LICENSE
```diff
@@ -1,3 +1,14 @@
+Membrane Live Compositor plugin is using Live Compositor that is licensed
+under Business Source License 1.1.
+
+See https://github.com/membraneframework/live_compositor/blob/master/LICENSE
+to learn more.
+
+-----------------------------------------------------------------------------
+
+Membrane Live Compositor plugin on its own is licensed under Apache-2.0 license
```
WojciechBarczynski, line 9: Membrane Live Compositor Plugin

LICENSE
```diff
@@ -1,3 +1,14 @@
...
+included below.
+
```
WojciechBarczynski, line 11: I'm not sure if I understand it. As a user, I would be confused and I wouldn't know if I can use plugin under Apache licence or not.

wkozyra95: If you are using it with LiveCompositor then no. If you are using it without LiveCompositor "on its own" then yes, but it does not make sense without it (which does not really make sense).

I think this statement is clear what license applies to this source code. Question "can use plugin under Apache licence or not" depends on the user.

Also semi-seriously, confusion is appropriate response when reading any license related stuff :D.

PR #197, opened 2024-07-03
Don't setup a ResourceGuard when the LiveCompositor is already started

nickdichev-firework:
When `server_setup` is already started it implies the client has started the LiveCompositor outside of our control. In this case the `server_pid` is nil and creates an ArgumentError when we try to `Process.exit`


WojciechBarczynski (approved):
Hi, thanks @nickdichev-firework for the contribution. Yep, we forgot about that 😅.

I'll merge this now. I'll release `v0.9.1` with this patch today, so you can bump your dependency :)

PR #199, opened 2024-07-12
Add support for requesting keyframes

WojciechBarczynski:


lib/membrane/live_compositor/api_client.ex
```diff
@@ -31,6 +33,11 @@ defmodule Membrane.LiveCompositor.ApiClient do
     {:get, "/status", nil} |> send_request(lc_port)
   end
 
+  @spec request_keyframe(:inet.port_number(), LiveCompositor.output_id()) :: request_result()
+  def request_keyframe(lc_port, output_id) do
+    {:post, "/api/output/#{output_id}/request_keyframe", %{}} |> send_request(lc_port)
```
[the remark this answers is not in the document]
wkozyra95: Also, I think `request.ex` would be a more appropriate place to add support for this, the same way as other requests already are. Keyframe requests could then be sent the same way registering inputs are.

lib/membrane/live_compositor/output_processor.ex
```diff
@@ -31,6 +46,28 @@ defmodule Membrane.LiveCompositor.VideoOutputProcessor do
...
   end
+
+  @impl true
+  def handle_event(:input, event, _ctx, state) do
+    {[event: {:output, event}], state}
+  end
+
+  @impl true
+  def handle_event(:output, %Membrane.KeyframeRequestEvent{}, _ctx, state) do
+    case ApiClient.request_keyframe(state.lc_port, state.output_id) do
```
wkozyra95, line 57: I'm not an expert, but I understand the "proper" way of handling sth like this in elixir would be to send a message to the parent process, and that process should send the actual request.

lib/membrane/live_compositor/output_processor.ex
```diff
@@ -4,6 +4,8 @@ defmodule Membrane.LiveCompositor.VideoOutputProcessor do
 
   use Membrane.Filter
 
+  require Membrane.Logger
+
```
wkozyra95, line 8:
```suggestion
```

lib/membrane/live_compositor/live_compositor.ex
```diff
@@ -767,4 +763,23 @@ defmodule Membrane.LiveCompositor do
   defp output_group_id(output_id) do
     "output_group_#{output_id}"
   end
+
+  @spec send_req(Request.t(), :inet.port_number()) :: ApiClient.request_result()
+  defp send_req(req, lc_port) do
```
wkozyra95, line 768:
```suggestion
  @spec handle_request(Request.t(), :inet.port_number()) :: ApiClient.request_result()
  defp handle_request(req, lc_port) do
```
plus I would pass state instead of just port

lib/membrane/live_compositor/live_compositor.ex
```diff
@@ -685,6 +674,13 @@ defmodule Membrane.LiveCompositor do
     {[notify_parent: {:output_registered, pad_ref, state.context}], state}
   end
 
+  @impl true
+  def handle_child_notification(:keyframe_request, {:output_processor, pad_id}, _ctx, state) do
```
wkozyra95, line 678: This is fine, but I would probably send `Request.KeyframeRequest` from the output processor

wkozyra95 (approved):

PR #203, opened 2024-08-12
Handle `OUTPUT_DONE` WebSocket message

WojciechBarczynski:


lib/membrane/live_compositor/lifecycle.ex
```diff
@@ -62,11 +62,15 @@ defmodule Membrane.LiveCompositor.Lifecycle do
   """
   @type input_eos :: :input_eos
 
-  @type notification :: {
-          output_registered()
-          | input_registered()
+  @type event ::
```
wkozyra95, line 65: If I remember correctly, you had strong objections to calling those events. I don't really see why you are adding new type, but i don't object to it either

wkozyra95 (approved):

PR #207, opened 2024-10-23
Spawn live compositor servers underneath a DynamicSupervisor

nickdichev-firework:
We recently transitioned our app from using a global live compositor which we spawned in our supervision tree and used the `:already_started` mode back to relying on the plugin to start the compositor. However, my test logs were being spammed with output from how it was being spawned.

I changed this to rely on the MuonTrap.Daemon which can be spawned under a DynamicSupervisor. This allows the logs to be captured in the normal way and let any logger config work as expected.

There's definitely a problem with the fact that I'm using String.to_atom here. Let me know what you think, probably we just need to shove that responsibility to the caller. Also, I use :transient mode since I'm pretty sure the pipeline will die if the compositor crashes, but let me know if that isn't true.

### DIFF
--- a/lib/membrane/live_compositor/application.ex
+++ b/lib/membrane/live_compositor/application.ex
@@ -1,0 +1,15 @@
+defmodule Membrane.LiveCompositor.Application do
+  @moduledoc false
+  use Application
+
+  @impl Application
+  def start(_type, _args) do
+    Supervisor.start_link(
+      [
+        {DynamicSupervisor, name: Membrane.LiveCompositor.ServerSupervisor}
+      ],
+      strategy: :one_for_one,
+      name: __MODULE__
+    )
+  end
+end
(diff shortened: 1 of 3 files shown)


lib/membrane/live_compositor/server_runner.ex
```diff
@@ -94,21 +94,22 @@ defmodule Membrane.LiveCompositor.ServerRunner do
...
+        "LIVE_COMPOSITOR_API_PORT" => to_string(lc_port),
+        "LIVE_COMPOSITOR_WEB_RENDERER_ENABLE" => "false"
+      })
+
+    spec =
+      Supervisor.child_spec(
+        {MuonTrap.Daemon, [bin_path, _args = [], [env: env, log_output: :info]]},
+        # TODO: make the caller pass this in, so they can use an atom pool if they're
+        # concerned about atom exhaustion
+        id: String.to_atom(instance_id),
```
varsill, line 108: Cannot we just do:
```suggestion
        id: instance_id,
```
?
I believe id might be just a `term()` : https://hexdocs.pm/elixir/1.12/Supervisor.html#t:child_spec/0

nickdichev-firework: Yes, I thought this had to be atom!

lib/membrane/live_compositor/application.ex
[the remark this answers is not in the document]
nickdichev-firework: Great idea, I updated it

varsill (approved):
Looks good to me 🥇

wkozyra95 (approved):

PR #211, opened 2024-12-05
Support Compositor running on another IP

FelonEkonom:


wkozyra95 (changes requested):
all tcp_sink and tcp_source elements are still connecting to localhost

lib/membrane/live_compositor/server_runner.ex
wkozyra95, line 1: I can't comment on the line, but you need to handle the option you added in ensure_server_started

varsill (approved):

lib/membrane/live_compositor/server_runner.ex
```diff
@@ -60,16 +60,45 @@ defmodule Membrane.LiveCompositor.ServerRunner do
...
+              ip_or_hostname |> to_charlist() |> get_host_by_name!()
+          end
+
         lc_address = {ip_address, opt.api_port}
         {:ok, lc_address, nil}
     end
   end
 
+  defp get_host_by_name!(hostname) do
+    with {:ok, ip} <- :inet.gethostbyname(hostname) do
```
wkozyra95, line 88: https://www.erlang.org/doc/apps/kernel/inet.html#gethostbyname/2
You can force ipv4 address with Family option

PR #214, opened 2025-02-18
Rename to `Smelter` and update to `0.4.0`

noituri:


mix.exs
```diff
@@ -1,12 +1,12 @@
-defmodule Membrane.LiveCompositor.Mixfile do
+defmodule Membrane.Smelter.Mixfile do
   use Mix.Project
 
-  @version "0.10.1"
-  @github_url "https://github.com/membraneframework/membrane_live_compositor_plugin"
+  @version "0.11.0"
```
wkozyra95, line 4:
```suggestion
  @version "0.10.1"
```
We will bump that in release commit, also it should be bumped at the same time as version in README

wkozyra95 (approved):

PR #222, opened 2026-04-13
Allow SMELTER_WEB_RENDERER_ENABLE to be set via environment variable

sarat1669:
The web renderer enable flag was hardcoded to "false". Now reads from System.get_env with "false" as default, allowing users to enable the web renderer by setting SMELTER_WEB_RENDERER_ENABLE=true.


wkozyra95:
Thanks for the contribution, this will work in some cases if you run against the smelter version you provided yourself, but by default, this plugin downloads a version of smelter compiled without web rendering support.

https://github.com/membraneframework/membrane_smelter_plugin/blob/master/lib/membrane/smelter/compile/download_compositor.ex#L71

